PR #7, opened 2020-11-12
Remove delta config for person reconciliation service

claire-lovisa:
Remove the delta rule triggering the person reconciliation service, because it causes issues in the frontend. 

The main problem is that a person newly created can be instantly deleted, inducing errors when trying to attach a mandataris to it.

### DIFF
--- a/config/delta/rules.js
+++ b/config/delta/rules.js
@@ -162,23 +162,6 @@ export default [
   },
   {
     match: {
-      predicate: {
-        type: 'uri',
-        value: 'http://www.w3.org/ns/adms#identifier'
-      }
-    },
-    callback: {
-      url: 'http://person-reconciliation/delta',
-      method: 'POST'
-    },
-    options: {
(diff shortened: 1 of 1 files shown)


MikiDi:
At a risk of not knowing what I'm talking about (saw this pass by in my notifications) ... If it 'd be that it needs to have only a couple more seconds of patience, you could bump the [`gracePeriod`](https://github.com/mu-semtech/delta-notifier#receiving-delta-notifications) to a minute or two (`120000`) ...

claire-lovisa:
Hey Michaël ! A risk always worth taking :D but I think in this case it will really help, in the sense that even if we wait a few seconds, the user will still be on a frontend page linked to a person's uuid that is not valid anymore so when he/she tries to refeesh it or do actions on it, the requests will fail. Running the service as a cron job will allow them to create all the persons, mandatarissen etc that they need, and it'll be sync during the night without them even noticing it :female_detective:

PR #17, opened 2021-01-08
Add kbonummer and check performance change

aatauil:
added following snippet to query, executed it 8 times to get a more accurate average but numbers are all over the place.end-result is +/- 7% increase but would rather add some margin and say it is somewhere between 5% to 9% increase 

 ```sql
  OPTIONAL {
    ?uri ext:kbonummer ?kbonummer.
  }
 ```

Original ( in milliseconds ) | modified ( in milliseconds )
-- | --
338 | 345
282 | 346
327 | 329
313 | 335
286 | 310
300 | 328
293 | 339
322 | 313

Original Average: 307

Modified Average: 330

Average difference: 23

Percentage increase: 6.96%

### DIFF
--- a/config/reports/bestuurseenhedenReport.js
+++ b/config/reports/bestuurseenhedenReport.js
@@ -16,10 +16,13 @@ export default {
       PREFIX ext: <http://mu.semte.ch/vocabularies/ext/>
       PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
       
-      select distinct ?uri ?name ?type ?province where {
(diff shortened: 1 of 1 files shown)


cecemel:
It seems KBO nummer is queried but not exported in the excel.

aatauil:
> It seems KBO nummer is queried but not exported in the excel.

My bad , focused to hard on the performance impact that I forgot to actually add it to csv. This has been added and tested and the kbonummer now gets added to the csv file

PR #20, opened 2021-01-13
Optimize/mandaten report

aatauil:
Filter mandataris, persoon & bestuursorgaan on  ?bestuursperiodeStart en check performance. Only performed the test twice since they take pretty long but I think the difference is pretty obvious

Original ( in minutes ) | modified ( in minutes )
-- | --
13.25 | 6.78
13.44 | 6.79

Original Average: 13.34

Modified Average: 6.78

Average difference: 6.56

Percentage decrease: 49.18%


cecemel:
Extracted the relevant commits. to merge them in master. closing here

PR #29, opened 2021-02-10
Clean migration names up

claire-lovisa:
### Updating migrations to rename them

/!\ Renaming migrations is a tricky and potentially risky process. We need to be sure that if some migrations run twice, it will not have any impact on existing data. 

#### Why ?

The need to rename migrations here is that the order of execution of the migrations wasn't correct and could induce weird states in the data when starting from a fresh new virtuoso container locally.

#### How ?

- Some migrations could not run twice without impacting the existing data (ones creating new resources with a random URI like `Aanvraag`). Fortunately, those migrations were only useful temporarily for subsidies that got submitted before the subsidy-management-service could generate those extra data itself. We could remove them safely.
- Some other migrations were adding data we need in the stack and not generated anywhere else, such as `SubsidiemaatregelAanbod` or new reeks. Fortunately again, those migrations insert or delete data with fixed URIs and data so they can run twice, inserting twice the same data and so not bringing any changes. We could rename them safely.
Some migrations were a mix of those two, this is why sometimes only parts of migrations were deleted and the file renamed.

#### How I tested it ?

- From a production dump, I restarted the migrations service and checked carefully the data related to subsidies to be sure they didn't change
- From an old production dump (07/2020, before subsidies), I ran all the migrations and checked the data generated when creating a subsidy


HugaertsDries:
Even though they have already run (under an other name that is) and their respective timestamps are in the past, from what I understand, these migrations will run on deployed/active environments. This wouldn't be an issue if the data hadn't changed.

But I think there are two migrations that need to be re-looked before re-running as the data has changed:

#### 1. [20210112165414-create-new-reeks](https://github.com/lblod/app-digitaal-loket/blob/5bc2f9f22be0352ca51f1289c9d07ca593871ab4/config/migrations/20201125120418-subsidy-management/20210112165414-create-new-reeks.sparql)
 
Will override changes made by [20210216160159-clean-up-labels-and-titles-of-reeksen](https://github.com/lblod/app-digitaal-loket/blob/40570ddd1fc136f2704df6fba616a197370dc6a8/config/migrations/20210205130451-new-subsidy-form-versioning/20210216160159-clean-up-labels-and-titles-of-reeksen.sparql). In this query the labels voor the concepts of subsidy were clean-up and renamed in preparation for the new form-versioning.

#### 2. [20210113154535-rename-config-file](https://github.com/lblod/app-digitaal-loket/blob/5bc2f9f22be0352ca51f1289c9d07ca593871ab4/config/migrations/20201125120418-subsidy-management/20210113154535-rename-config-file.sparql)

For the most part this query seems ok, except for the first 2 queries within. At this point in time the new form-versioning is introduced. For this [20210205205722-remove-old-form-directories-from-the-store](https://github.com/lblod/app-digitaal-loket/blob/e32ca033d8251209c5accb18ddd0b687a67ebfcc/config/migrations/20210205130451-new-subsidy-form-versioning/20210205205722-remove-old-form-directories-from-the-store.sparql) cleaned-up the old form-directories. Correct me if I'm wrong but I'm afraid the INSERT DATA will add back a triple that should be removed.

For me if those two migrations have been considered and updated, the PR is ready to be merged! :)

HugaertsDries (commented):
Even though they have already run (under an other name that is) and their respective timestamps are in the past, from what I understand, these migrations will run on deployed/active environments. This wouldn't be an issue if the data hadn't changed.

But I think there are two migrations that need to be re-looked before re-running as they affected data that has seen changes:

#### 1. [20210112165414-create-new-reeks](https://github.com/lblod/app-digitaal-loket/blob/5bc2f9f22be0352ca51f1289c9d07ca593871ab4/config/migrations/20201125120418-subsidy-management/20210112165414-create-new-reeks.sparql)
 
Will override changes made by [20210216160159-clean-up-labels-and-titles-of-reeksen](https://github.com/lblod/app-digitaal-loket/blob/40570ddd1fc136f2704df6fba616a197370dc6a8/config/migrations/20210205130451-new-subsidy-form-versioning/20210216160159-clean-up-labels-and-titles-of-reeksen.sparql). In this query the labels voor the concepts of subsidy were cleaned-up and renamed in preparation for the new form-versioning.

#### 2. [20210113154535-rename-config-file](https://github.com/lblod/app-digitaal-loket/blob/5bc2f9f22be0352ca51f1289c9d07ca593871ab4/config/migrations/20201125120418-subsidy-management/20210113154535-rename-config-file.sparql)

For the most part this query seems ok, except for the first 2 queries within. At this point in time the new form-versioning is introduced. For this [20210205205722-remove-old-form-directories-from-the-store](https://github.com/lblod/app-digitaal-loket/blob/e32ca033d8251209c5accb18ddd0b687a67ebfcc/config/migrations/20210205130451-new-subsidy-form-versioning/20210205205722-remove-old-form-directories-from-the-store.sparql) cleaned-up the old form-directories. Correct me if I'm wrong but I'm afraid the INSERT DATA will add back a triple that should be removed.

For me if those two migrations have been considered and updated, the PR is ready to be merged! :)

claire-lovisa:
@HugaertsDries Thanks for catching those :D It allowed me to also see I needed to update `inScheme` when creating reeks. After a lot of testing, it should be good to go ! I'll merge it to dev.

PR #31, opened 2021-02-17
Add download-url-warning-service to the stack

claire-lovisa:


docker-compose.yml
```diff
@@ -409,3 +409,9 @@ services:
       - "logging=true"
     restart: always
     logging: *default-logging
+  download-url-warning:
+    image: lblod/download-url-warning-service:0.0.2
```
HugaertsDries, line 413: any reason we start from `0.0.2`? Usually we start a service off with `0.1.0`

claire-lovisa: mmh I think just me not being aware of that convention haha, will fix that

PR #32, opened 2021-02-22
Add new form version with missing hidden inputs

claire-lovisa:
Some hidden inputs necessary to conditionally display other fields were missing on the Wijziging-ocmw-vereniging form.


HugaertsDries (approved):
Checked it out, forms still work fine! Also tested as you suggested, the classification codes are in the source data! So an approval from me! :+1:

PR #42, opened 2021-04-06
Feature/subsidie flow prepare api

cecemel:
Makes the migration + api from subsidies old to new flow.  ( it is also probably a long standing branch, so backing up fundamental data changes makes sense)

- to run/test this, make you make a backup of your data. I assume it is fine, but since it is a tedious bookkeeping operation, I don't exclude a priori mistakes
- Still left is attaching the correct application form to SMC (subsidiemaatregelconsumptie)
- also needs SMC statuses, we need to come up with status there
- type attached to SPS (subsidieprocedurestap) needs further refinment


cecemel:
all right here we go.

PR #48, opened 2021-05-03
Cleanup functionarissen

claire-lovisa:
This PR:
- Archives wrongful functionarissen
- Archives duplicated functionarissen
- Updates the export configuration so that it gets propagated to LDB


aatauil:
All tested and working as should, thankss

PR #52, opened 2021-05-18
Updated and new subsidy reports

claire-lovisa:


aatauil (approved):
Everything looks good, new form get included in the report which is perfect :) all columns with correct data present in report. Great job !

PR #53, opened 2021-05-28
Add and fix reports

claire-lovisa:
**Added**
- Report about elected persons -> fieds asked where `"personURI"; "name"; "familyName";"listName";"cityName";"rrnr"`

**Updated**
- Report mar7300 -> issue with optional values blocking the process + a little rework
- Report mandatarissen with empty person -> same, fix optional values causing issues
- Report same RRN -> updated prefixes and compare actual RRNs instead of RRN URI's


config/reports/sameRrnReport.js
```diff
@@ -16,10 +16,14 @@ export default {
       PREFIX adms: <http://www.w3.org/ns/adms#>
 
       SELECT DISTINCT ?person1 ?person2 WHERE {
-        ?person1 a person:Person.
-        ?person2 a person:Person.
-        ?person1 adms:identifier/skos:notation ?rrn.
-        ?person2 adms:identifier/skos:notation ?rrn.
+        GRAPH ?g {
+          ?person1 a person:Person ;
```
cecemel, line 20: why do we need this?

cecemel: It seems it gives  different results:

```
      PREFIX person: <http://www.w3.org/ns/person#>
      PREFIX adms: <http://www.w3.org/ns/adms#>
      PREFIX foaf: <http://xmlns.com/foaf/0.1/>
      SELECT DISTINCT ?person1  ?person2 WHERE {

        ?person1 a person:Person.
        ?person2 a person:Person.
        ?person1 adms:identifier/skos:notation ?rrn.
        ?person2 adms:identifier/skos:notation ?rrn.
        FILTER (?person1 != ?person2)
      }
```

cecemel: I am also wondering why they haven't been picked up by the clean up service

claire-lovisa: @cecemel oh good catch, actually I reused the query of the [person-reconciliation-service](https://github.com/lblod/person-reconciliation-service/blob/fcbd3ccdde1d6174ec4813e7c452ea9e94a68843/support.js#L64-L89) to fix the report, so it's normal that the service doesn't pick up anything as the result when specifying the graph is empty.
What that means is that in some cases, the duplicates reside in different graphs, and if we want to merge them we should extend (or update the query) of the reconciliation service

PR #54, opened 2021-06-02
Add detailed report on contact tracing subsidies

claire-lovisa:


aatauil (changes requested):
I tested with both old submissions and new ones and they both work fine. There are minor issues I found: when the field 'uitleggenVanMaatrgelen' is checked, it will say "nee" instead of "ja" in the report. Same goes for "financieleHulpvragenDoorgegevenNaarDeJuisteInstanties". The rest is accurate :)

PR #59, opened 2021-06-21
Remove some mandates for gemeente Herent

claire-lovisa:


aatauil (approved):

aatauil:
Tested and migration did its job. Changed merging branch to development instead of master

PR #67, opened 2021-07-06
Add new semantic form version with rapportjaar validation

claire-lovisa:


aatauil (approved):
Error message showing when entering invalid year in rapportjaar input :) thanks for the changes

PR #69, opened 2021-07-15
Upate export filter of submissions

claire-lovisa:
This prevents the export/import path from crashing. It's a temporary fix, that will be definitely taken care of when the deltas will be fully functional.


aatauil (approved):
File generated and data seems correct, thanks!

PR #80, opened 2021-08-18
add hidden validation field tables + cleanup

aatauil:
- estimatedCost table was missing hidden input that validates the table. Cleanup file to be more consistent with others. Tested against pre-existing forms as wel as new ones

### DIFF
--- a/config/subsidy-application-management/forms/bicycle-infrastructure/request/versions/20210504170646/form.ttl
+++ b/config/subsidy-application-management/forms/bicycle-infrastructure/request/versions/20210504170646/form.ttl
@@ -14,6 +14,18 @@
 @prefix ext: <http://mu.semte.ch/vocabularies/ext/>.
 @prefix bicycleInfrastructure: <http://lblod.data.gift/vocabularies/subsidie/bicycle-infrastructure#>.
 
+
+##########################################################
+# Form definition
+##########################################################
+
+form:3ff36255-f108-4527-990c-6d7896f4b989 a form:Form ;
+    mu:uuid "3ff36255-f108-4527-990c-6d7896f4b989" ;
+    form:hasFieldGroup fieldGroups:1f3cf92c-aa5f-4708-85c6-77aabbc8989f .
+
+fieldGroups:1f3cf92c-aa5f-4708-85c6-77aabbc8989f a form:FieldGroup ;
+    mu:uuid "1f3cf92c-aa5f-4708-85c6-77aabbc8989f" .
+
 ##########################################################
 # Property groups
 ##########################################################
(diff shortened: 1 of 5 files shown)


aatauil:
@claire-lovisa @cecemel  Created seperate folder for new version. Added migration to link new form.ttl version.

The subsidy-application-management-service from what I found in the code does not use the 'latest' dct:source explicitly [SEE QUERY](https://github.com/lblod/subsidy-applications-management-service/blob/1cbd68ff323d3f1a625cb68ffae39fa1b6d62855/lib/services/semantic-form-management.js#L349-L356).  So I am kinda afraid that it uses the latest version of the form by coincidence as the service just takes the result in position [0]. 

I dont know how virtuoso sorts things by default but even though its correct now, it might not be in the future when adding new versions

aatauil:
Seperated into new version folder and created migration to add new dct:source. Now it is using the latest version in this case but I am not sure if this behaviour was intended see [query](https://github.com/lblod/subsidy-applications-management-service/blob/6fe06ebb6325b41e1a7041ff464ab74d6ceb2c57/lib/services/semantic-form-management.js#L349-L356).

config/subsidy-application-management/forms/bicycle-infrastructure/request/versions/20210504170646/form.ttl
```diff
@@ -317,4 +317,4 @@ fieldGroups:1f3cf92c-aa5f-4708-85c6-77aabbc8989f form:hasField fields:94f3ab67-e
 fieldGroups:1f3cf92c-aa5f-4708-85c6-77aabbc8989f form:hasField fields:39b5f8a3-f163-4198-bf0d-82efeb2c0254 .
 fieldGroups:1f3cf92c-aa5f-4708-85c6-77aabbc8989f form:hasField fields:d91b1cc3-f2a1-4704-884c-e89f612cba33 .
 fieldGroups:1f3cf92c-aa5f-4708-85c6-77aabbc8989f form:hasField fields:1f17de22-e0c9-4166-ad32-a0d1dc11e375 .
-fieldGroups:1f3cf92c-aa5f-4708-85c6-77aabbc8989f form:hasField fields:2d154583-7e39-4bbf-9a53-ea7fd95e2c96 .
```
cecemel, line 320: This is an existing form, already submitted. I suppose we don't want to edit this one?

cecemel (changes requested):
I am still able to submit this. After restarting and running migrations and looking the correct new step is taken into account
image
Also, since it is the same component, in the first step, shouldnt we make another form here?

config/subsidy-application-management/forms/bicycle-infrastructure/request/versions/20210504170646/form.ttl
aatauil: Github marked it as 'changed' but there it a little red mark at the end saying: "No new line at end of file" when you hover over it so nothing really changed

PR #84, opened 2021-09-06
Add one report per step of bike subsidy

claire-lovisa:


config/reports/fietsSubsidieRequestsReport.js
```diff
@@ -0,0 +1,71 @@
...
+        ?bestuur m8g:playsRole ?participation ;
+          skos:prefLabel ?bestuurseenheid .
+
+        OPTIONAL { ?applicationForm <http://schema.org/bankAccount>/<http://schema.org/identifier> ?iban . }
+
+        FILTER EXISTS {
+          ?applicationForm dct:isPartOf ?step .
+          ?step dct:references <http://data.lblod.info/id/subsidieprocedurestappen/52f0b7dd244e42e0cda83804508e2e89d94ed098f3df8b4f9913a14f2be2423d> .
+        }
+      }
```
cecemel, line 40: @claire-lovisa do we want also to add
```
        FILTER NOT EXISTS {
          ?applicationForm adms:status <http://lblod.data.gift/concepts/6b7ae118-4653-48f2-9d9a-4712f8c30da9>. # new
        }``

File: config/reports/fietsSubsidieRequestsReport.js [hunk 2]
```diff
@@ -0,0 +1,71 @@
...
+          dct:source ?applicationForm .
+        ?bestuur m8g:playsRole ?participation ;
+          skos:prefLabel ?bestuurseenheid .
+
+        OPTIONAL { ?applicationForm <http://schema.org/bankAccount>/<http://schema.org/identifier> ?iban . }
+
+        FILTER EXISTS {
+          ?applicationForm dct:isPartOf ?step .
+          ?step dct:references <http://data.lblod.info/id/subsidieprocedurestappen/52f0b7dd244e42e0cda83804508e2e89d94ed098f3df8b4f9913a14f2be2423d> .
+        }
```
cecemel, line 39: would it make sense to add:
```
        FILTER NOT EXISTS {
          ?applicationForm adms:status <http://lblod.data.gift/concepts/6b7ae118-4653-48f2-9d9a-4712f8c30da9>. # new
        }```

cecemel (changes requested):

config/reports/fietsSubsidieRequestsReport.js
claire-lovisa: It would I think, maybe not on step 1 because they might want to see if the bestuur has at least started something, but on step 2 indeed it'll just be blank

cecemel: It is for step 2 right? Because, now as soon as they click on the next step, even if the step 1 is not ok, they are taken in the report.

claire-lovisa: Yes. What I meant is that now, a step 1 will also be "New" until the first time it's saved, but we might want to keep it in the report for step 1.
Step 2 --> Only "Concept" ones are interesting

claire-lovisa:
@cecemel Pushed an update :) Can you re-check and tell me if ok to merge ? Thanks !

PR #98, opened 2021-10-18
Add migration to revert step 2 of subsidy

claire-lovisa:


HugaertsDries (approved):
@claire-lovisa Maybe we could somehow bundle this with my removal of copy consumptions and release a hotfix later this week?

claire-lovisa:
@HugaertsDries Yes good idea to bundle it ! We can also deploy #97 at the same time ?

PR #112, opened 2021-11-19
Subsidy: GZG Oproep 4

HugaertsDries:


claire-lovisa (approved):
Works as expected :)
Small note, but it seems to be the case for all the subsidies of this type : when submitting, the SMC stays in the "Actief" state, not sure if it's expected or not

PR #117, opened 2022-01-02
Feature/cleanup files from public

cecemel:


cecemel:
This PR bundles the exit from files from the public graph. Once the full cut-over has been done in code, so no write-access anymore of files to public, we perform a full CRUD testing of the app. (Before cut-over, this  isn't worth the effort, as the tests might be flawed,  due to the fact that files might still be stored in the public graph)

PR #122, opened 2022-01-19
Add the "Vaststelling gemeentelijk beleidskader (gemeentewegen) form

Windvis:
Related PR: https://github.com/lblod/manage-submission-form-tooling/pull/8

This also includes a small html fix for some help text contents to make it consistent with the version for the new form.


config/migrations/2022/20220119141600-new-form/20220119141700-add-new-besluit-type.ttl
```diff
@@ -0,0 +1,15 @@
+INSERT DATA {
```
Windvis, line 1: I had some issues with this migration and I still haven't figured it out. After the migration ran (the migration service marked it as done) the data wasn't actually added in the database which meant the form option wasn't visible. After copying this query in Virtuoso's SPARQL interface it did show work so it's not the query itself.

HugaertsDries: @Windvis does this issue still persist? I think it's due to the file extension being defined as `.ttl` instead off `.sparql`. It is weird that the migrations service wouldn't complain, that I do admit.

Windvis: Doh, that will be it for sure, thanks. I haven't retested it to be honest, I didn't create a backup of a "clean db" and running the migrations from 0 takes a very long time.

HugaertsDries: :arrow_up: yup, that's seems to do the trick!

HugaertsDries (approved):
Form works as adventist, thumb-up from me!

config/migrations/2022/20220119141600-new-form/20220119141700-add-new-besluit-type.ttl
Windvis: Thanks!

PR #136, opened 2022-02-14
Add the new `POST /files` endpoint

Windvis:
Both `@lblod/ember-vo-mu-file-upload` and appuniversum use the `/files` endpoint to upload new files (by default).

### DIFF
--- a/config/dispatcher/dispatcher.ex
+++ b/config/dispatcher/dispatcher.ex
@@ -112,6 +112,11 @@ defmodule Dispatcher do
   patch "/files/*path" do
     Proxy.forward conn, path, "http://resource/files/"
   end
+  post "/files/*path" do
+    Proxy.forward conn, path, "http://file/files/"
+  end
+  # TODO: find all usage of this endpoint and replace it with `POST /files`
+  # This is kept to maintain compatibility with code that uses the "old" endpoint.
   post "/file-service/files/*path" do
     Proxy.forward conn, path, "http://file/files/"
   end


aatauil:
I think quite a lot of services use this endpoint. I dont really know if it is worth changing it to fit appuniversum instead of the other way around. Will do some analysis and paste a summary on my findings

Windvis:
This came out of a discussion with Felix and Dries. Other apps are switching to the `/files` endpoint as well (which is why it was made the new default in all the upload components).

I didn't do any research to see were it's used though so it's not bad that we know where this endpoint is used :+1: 

Supporting both endpoints was the quickest way to unblock updates without having to change everything at once.

aatauil:
Oke so summary of my findings. I am a big fan of consistentie and if we want consistentie across the board then this change will probably have an impact on other projects too, but it doesnt have to since everyone can just go their way.

*Dispatcher of other apps using the /file-service/files post route*
https://github.com/lblod/app-complaint-form
https://github.com/lblod/app-poc-form-builder
https://github.com/lblod/app-meldingsplichtige-api
https://github.com/lblod/app-toevla

Not sure if the post route is actually used in some of these since the their frontend doesnt reference any routes for it. Except toevla does have a frontend app that does
https://github.com/lblod/frontend-toevla-data-entry

If we look at it in isolation then removing the old file-service/files post route shouldnt change anything as I didnt find any other occurences in the frontend code or microservices that depend on it. Needs to be put in grace periode in dev environment first to be 100% sure

aatauil (approved):

PR #144, opened 2022-02-28
Feature/update fietssubsidies steps

cecemel:


aatauil:
Looking great! Already did an intermediate review. Might be good to keep this PR alive until after we also implement the table itself. Noticed a small detail when actually sending the form. Conditional fields get hidden. I think this is a known issue? 

Created an issue so we dont forget to look into it as it is a lot more prominent in samenleving subsidy then other ones.
https://github.com/lblod/ember-submission-form-fields/issues/43

PR #149, opened 2022-03-14
Feature/climate subsidy oproep 2

cecemel:


Windvis:
I pushed some commits to update the description and add properties for the custom action rows to the form config.

PR #151, opened 2022-03-21
Feature/subsidy plan samenleven Table

aatauil:
Added table to plan samenleven subsidy
- The form.ttl & form.json for plan samenleven subsidy have a new property group 'objectives' and to that property group a Table has been added as field.
- subsidy rule migration has been added that contains all rows and values for the table similar to climate subsidy table
note that the values are hardcoded in submission-form-fields


aatauil:
to be tested together with ember-submission-form-fields updates see https://github.com/lblod/ember-submission-form-fields/pull/46

PR #163, opened 2022-05-05
Feature/ipdc lpdc

bdevloed:


config/resources/lpdc.json
```diff
@@ -0,0 +1,643 @@
...
+          "target": "conceptual-public-services",
+          "predicate": "m8g:hasContactPoint",
+          "cardinality": "many",
+          "inverse": true
+        }
+      },
+      "new-resource-base": "http://data.lblod.info/id/contact-punten/"
+    },
+    "documents": {
+      "class": "dcat:Distribution",
```
cecemel, line 537: dcat:distribution is a bit suprising here

config/resources/lpdc.json
```diff
@@ -0,0 +1,644 @@
...
+          "target": "identifiers",
+          "predicate": "dct:identifier",
+          "cardinality": "one"
+        },
+        "related": {
+          "target": "conceptual-public-services",
+          "predicate": "dct:relation",
+          "cardinality": "many"
+        },
+        "hasInstance": {
```
Windvis, line 83: I think some of these relationship names are a bit weird. They will be used in the frontend as defined here, so for example we would have to use `conceptualPublicServiceRecord.hasInstance` to receive a list of services which isn't intuitive when reading the code 😄. `instances` might be a more intuitive name, maybe?

I would try to give all relationships a more logical sounding name, especially the ones with "many" cardinality.

config/migrations/2022/20220506-lpdc-ipdc/20220506-public-services-code-lists.graph
```diff
@@ -0,0 +1 @@
+http://mu.semte.ch/graphs/public
```
Windvis, line 1: The timestamp in the filename / folder name doesn't seem to include time information so it doesn't get sorted properly compared to the other migrations.

config/dispatcher/dispatcher.ex
```diff
@@ -580,6 +580,97 @@ defmodule Dispatcher do
     Proxy.forward conn, path, "http://cache/distributions/"
   end
 
+  #################################################################
+  # Public Services - LPDC-IPDC-PoC
+  #################################################################
+  match "/conceptual-public-services/*path" do
+    Proxy.forward conn, path, "http://resource/conceptual-public-services/"
```
Windvis, line 587: I'm not 100% sure but I assume most of these can go through the cache service to improve performance?

config/dispatcher/dispatcher.ex
```diff
@@ -580,6 +580,97 @@ defmodule Dispatcher do
...
+  match "/documents/*path" do
+    Proxy.forward conn, path, "http://resource/documents/"
+  end
+
+  match "/locations/*path" do
+    Proxy.forward conn, path, "http://resource/locations/"
+  end
+
+  match "/lifecycle-statuses/*path" do
+    Proxy.forward conn, path, "http://resource/lifecycleStatuses/"
```
Windvis, line 655: I think the resource names need to be dasherized (or simply match the version in the lpdc.json file, I'm not sure).

config/resources/lpdc.json
```diff
@@ -0,0 +1,644 @@
...
+        "label": {
+          "type": "string",
+          "predicate": "skos:prefLabel"
+        }
+      },
+      "new-resource-base": "http://lblod.data.gift/concepts/"
+    },
+    "service-types": {
+      "TODO": "COFOG Code list? cfr dienstencataloog AP",
+      "class": "code:ServiceType",
```
Windvis, line 624: It seems this predicate isn't the one that's used in the migration files, so it doesn't return any data.

`http://lblod.data.gift/vocabularies/lpdc-ipdc/IPDCServiceType`

config/resources/lpdc.json
```diff
@@ -0,0 +1,644 @@
...
+        "label": {
+          "type": "string",
+          "predicate": "skos:prefLabel"
+        }
+      },
+      "new-resource-base": "http://lblod.data.gift/concepts/"
+    },
+    "sectors": {
+      "TODO" : "NACE codes? cfr dienstencataloog AP",
+      "class": "code:Sector",
```
Windvis, line 593: Could it be that there is no data for this yet?

config/resources/lpdc.json
bdevloed: Sure, I can drop the `has*` from the json api. It's now literally taken from the rdf properties.

config/resources/lpdc.json
Windvis: Is there any data for it in the Sharepoint files, or do we need to request that somewhere? It's used in some table columns and filters so it has an impact on the frontend.

config/resources/lpdc.json
```diff
@@ -0,0 +1,653 @@
...
+          "cardinality": "many"
+        }
+      },
+      "features": [
+        "include-uri"
+      ],
+      "new-resource-base": "http://data.lblod.info/id/conceptual-public-service/"
+    },
+    "identifiers": {
+      "class": "adms:Identifier",
```
Windvis, line 171: It seems that this is defined as `adms:identifier` in the migration files:

```
<http://data.lblod.info/id/lpdc-ipdc/00411460f7c92d2124a67ea0f4cb5f85> a <http://www.w3.org/ns/adms#identifier> ;
    skos:notation 363 .
```

I think that needs to be capitalized to match the config here?

Windvis: Wait, that's the "structured identifier part" nevermind 😄.

config/dispatcher/dispatcher.ex
bdevloed: Indeed, using cache now

config/resources/lpdc.json
bdevloed: Correct, fixed it, also changed other codelists.
ToDo is about linking this to an existing external code list (beyond PoC)

config/resources/lpdc.json
bdevloed: No data in the sharepoint files

bdevloed: Let me know if something should be mocked

config/resources/lpdc.json
bdevloed: It's correct that the class is with a capital and this was wrong in the data. Well spotted! I've pushed a migration to change this.
Currently we only have one (made up) conceptual service about eID. All other services are "concrete".
It is indeed used for the PID of the services.

bdevloed:
Thanks for the feedback!
The mock login used for lpdc-ipdc:
`POST mock/sessions/`
```
{
    "data": {
        "relationships": {
            "account": {
                "data": {
                    "id": "9de34754d3d9c601eef6680e2ec14599",
                    "type": "accounts"
                }
            },
            "group": {
                "data": {
                    "id": "50ede269-5c49-41cd-be68-ac135314019c",
                    "type": "groups"
                }
            }
        },
        "type": "sessions"
    }
}
```
There's a known issue that some resources might return more data then expected, such as the identifier, because this class is used elsewhere as well. However, these are not linked to the (conceptual) public services.

PR #166, opened 2022-05-18
Report Ukraine Subsidy

FangioD:
Test:
- Make sure some Ukraine subsidies are available
- docker-compose exec report-generation /bin/bash
- wget -O- --header='Content-Type:application/json' --post-data='{"data":{"attributes":{"reportName":"ukraineSubsidyOproepOneReport"}}}' 'http://localhost/reports'

One possible enhancement => reeksStart is 2022-3-14 instead of 2022-03-14 as shown in example report

### DIFF
--- a/config/reports/ukraineSubsidyOproep1Report.js
+++ b/config/reports/ukraineSubsidyOproep1Report.js
@@ -2,7 +2,7 @@ import { generateReportFromData, batchedQuery } from "../helpers.js";
 import { getSafeValue } from "./util/report-helpers";
 
 export default {
-  cronPattern: "0 0 0 * * *",
+  cronPattern: "0 30 23 * * *",
   name: "ukraineSubsidyOproepOneReport",
   execute: async () => {
     const reportData = {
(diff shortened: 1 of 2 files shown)


config/reports/ukraineSubsidyOproep1Report.js
```diff
@@ -0,0 +1,137 @@
+import { generateReportFromData, batchedQuery } from "../helpers.js";
+import { getSafeValue } from "./util/report-helpers";
+
+export default {
+  cronPattern: "0 0 0 * * *",
```
aatauil, line 5: We might have preferred another moment as all GzG rapports also get generated at midnight. We prefer to spread the load a bit. perhaps this might be better `0 30 23 * * *` 23h30

aatauil:
Rest looks good :) thanks for the work

config/reports/ukraineSubsidyOproep1Report.js
FangioD: Did not know this, I will change it.

aatauil (approved):

PR #165, opened 2022-05-17
Show concept forms + contact fields

aatauil:
- also show forms in concept
- most important is contact fields so those are wrapped in OPTIONAL

To test:
- create some noodopvang subsidy forms
- save to concept

- `docker-compose exec report-generation /bin/bash`
- ```wget -O- --header='Content-Type:application/json' --post-data='{"data":{"attributes":{"reportName":"noodopvangSubsidiesReport"}}}' 'http://localhost/reports'```
- check for report in data/files

PR #167, opened 2022-05-18
Hotfix/show concepts noodopvang

cecemel:


cecemel:
See also https://github.com/lblod/app-digitaal-loket/pull/165

PR #174, opened 2022-06-10
create form.json & form.ttl for eigenschappen tab

aatauil:
- Added fields for eigenschappen 
- checked for working crud

note: some fields are not in use yet. They have been appended with "TODO"

### DIFF
--- a/config/lpdc-management/characteristics/form.json
+++ b/config/lpdc-management/characteristics/form.json
@@ -6,7 +6,8 @@
       "http://data.lblod.info/concept-schemes/id/bf7fcb58-7258-42f7-b7bf-628cf4701842",
       "http://data.lblod.info/concept-schemes/id/1c648cd2-29fc-4ee1-a4e2-a82f57ab592d",
       "http://data.lblod.info/concept-schemes/id/9357bf70-ae33-40e6-bd2a-80e842f92d9a",
-      "http://data.lblod.info/concept-schemes/id/bf7fcb58-7258-42f7-b7bf-628cf4701842"
+      "http://data.lblod.info/concept-schemes/id/bf7fcb58-7258-42f7-b7bf-628cf4701842",
+      "http://data.lblod.info/concept-schemes/id/b790e1d2-fd96-4476-8c07-c76a017fed09"
     ]
   }
 }
--- a/config/lpdc-management/characteristics/form.json
+++ b/config/lpdc-management/characteristics/form.json
@@ -1,63 +1,12 @@
 {
-  "source": {
-    "prefixes": [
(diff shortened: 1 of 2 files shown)


Windvis:
I did a quick test and it does seem that values are being persisted, but I noticed that even though I changed the status, the "pill" in the top bar doesn't match the select value even after a page refresh. Any idea what might cause this? Could that be the cache service?

aatauil:
@Windvis Yeah this is because the values in the form and the one you get in the toolbar are seperate. Form comes from ttl & lpdc-management service while the toolbar values come from resources. (90% sure of this)

aatauil:
I dont think there is a ,non dirty , way to sync this. at least not 'live' so when a user changes the status and it reflecting without saving. When saving we could force a refresh for that routes model

Windvis:
@aatauil [We already reload the record data after a save](https://github.com/lblod/frontend-loket/blob/feature/lpdc-module/app/components/public-services/details-form.js#L87) but the backend returns the old data again. I think we need to bust the cache for that specific record somehow, but I'm not sure that's possible from a custom micro service?

Maybe we can just ditch the cache for these resource types for now? As a quick fix.

aatauil:
@Windvis yes, we decided for GET request to public-services to point directly to resources
see: https://github.com/lblod/app-digitaal-loket/commit/d77232f8ab0547e9daaa08a2a3b033477615868f
will test it right away

aatauil:
@Windvis Discoverd the issue (frontend-related) So I noticed that indeed the frontend re-fetches the data but we missed something. In the original findRecord we use 'include' for the type and status. When you reload() the publicService model it will not re-fetch those relationships. 

We need to explicitly also reload those relationships
publicService.status.reload()
publicService.type.reload()
publicService.reload()

(unless ember-data has an easier way but didnt find it in the docs)

relevant files
- router: https://github.com/lblod/frontend-loket/blob/chore/delete-public-service/app/routes/public-services/details.js
- save function: https://github.com/lblod/frontend-loket/blob/chore/delete-public-service/app/components/public-services/details-form.js#L87

Windvis:
@aatauil Ahh, good catch. I pushed a fix for that in this pr: https://github.com/lblod/frontend-loket/pull/124/commits/bbf058cf33d1592daff44a30366fa0b6a1e5e608 After that the status seems to properly update 👍.

I'm not a fan of manually reloading the relationships in the save method since that means we have to maintain the list in 2 places. Reloading the model hook of the details route seems like a better way to do it, but that would also show the loading template again and we can't tell Ember not to do that. So I went with a util instead. It could be changed to a service so we don't need to pass the store, but I thought I would keep it simple 😄.

PR #176, opened 2022-06-16
Prefilling climate subsidie call 2

FangioD:
Creating a climate subsidy call 2 will prefill contact information with latest sent climate subsidy or latest concept if none are sent.


FangioD:
To test, generate a climate subsidie oproep 1 (2021) with testMode=true. Fill in the fields of indienen pact (step 1).
Generate a climate subsidie oproep 2 (2022) and the fields should be prefilled, check if you can open and add files and it is possible to sent the form.

PR #179, opened 2022-06-24
Add migration to clean up extra uuids

claire-lovisa:
See DL-3878 for context

I decided to remove uuids that didn't match with the uuid in the URI (instead of randomly) because it didn't take more time to write and it'll be easier for devs later one to not get confused. It also made it easier to ensure we'll still have at least one uuid left.

To double check my results, I used the following queries:
- Data with two uuids
```
PREFIX mu: <http://mu.semte.ch/vocabularies/core/>

SELECT DISTINCT * where {
  ?s a ?type ;
    mu:uuid ?uuid1, ?uuid2 .

  FILTER (?type IN (
    <http://www.w3.org/ns/person#Person>,
    <http://www.w3.org/ns/adms#Identifier>,
    <http://data.vlaanderen.be/ns/persoon#Geboorte>
  ))

  FILTER (?uuid1 < ?uuid2)
}
```
- Data with no uuid
```
PREFIX mu: <http://mu.semte.ch/vocabularies/core/>

SELECT DISTINCT * where {
  ?s a ?type .

  FILTER (?type IN (
    <http://www.w3.org/ns/person#Person>,
    <http://www.w3.org/ns/adms#Identifier>,
    <http://data.vlaanderen.be/ns/persoon#Geboorte>
  ))

FILTER NOT EXISTS { ?s mu:uuid ?uuid }
}
ORDER BY ?s
```

### DIFF
--- a/config/migrations/2022/20220624095713-remove-extra-uuids.sparql
+++ b/config/migrations/2022/20220624095713-remove-extra-uuids.sparql
@@ -1,0 +1,83 @@
+# Due to some error while reconciliating persons a long time ago, we ended up having resources with 2 uuids
+# In those queries we clean them by :
+#  - Getting the resources with at least two different uuids
+#  - Removing the uuids that don't match with the one in the URI (for ease of writing the query + clarity)
+
+PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
+
+DELETE {
+  GRAPH ?g {
+    ?s mu:uuid ?uuid2 .
+  }
+} WHERE {
+  GRAPH ?g {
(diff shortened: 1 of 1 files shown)


cecemel (changes requested):
@claire-lovisa looking good, we just have to filter out the publication graphs I think (else no deltas will be published)

claire-lovisa:
I filtered out the producer graphs where it was happening, better now ?

claire-lovisa:
@cecemel aha, didn't see it in the data but better be safe than sorry : fix pushed !

PR #185, opened 2022-07-20
Changed mandataris rangorde type to string

benjay10:
This PR changes the type of the mandataris rangorde to a simple string instead of a `language-string`.

Because of a bug in Virtuoso, strings with language tag can, in some cases, not be removed from the triplestore. See https://github.com/openlink/virtuoso-opensource/issues/1055. To workaround this issue, these particular language tags can be removed with the help of a migration, and served correctly after a change in the relevant resource configuration. Some changes in the frontend are on the way too (see https://github.com/lblod/frontend-loket/pull/138).


cecemel:
@benjay10 thanks!

PR #187, opened 2022-07-26
Added migration query to remove timezones in dates of mandatarissen

benjay10:
This migration query removes the time and timezone information from start and end dates of mandatarissen. This is because these dates are only used to refer to a specific day, and not a specific time, but because the time is often entered in the wrong timezone, the date is sometimes interpreted wrongly.
With timezone removed from the date, the date is always interpreted the way the frontend wants to. Work is on its way to make the frontend always enter the correct date and time.


benjay10:
Won't be changed; discussed with @cecemel and @Windvis.

PR #203, opened 2022-08-19
Move the concept-scheme config into a dedicated file

Windvis:
This removes the duplicated concept-scheme config and moves it into a dedicated file. 

~~This targets #201 since that removes the config that depended on the concept-scheme.json file.~~ targets dev now.


cecemel:
@Windvis fyi, rebased, force pushed and had to fix a little issue. merging!

PR #204, opened 2022-08-23
Added locations codelist

cecemel:
Locations code list: it has it's lpdcExt: namespace, since it might still wobble a bit.


Asergey91 (approved):
works!

PR #215, opened 2022-09-01
Erediensten data import

claire-lovisa:
DL-4121

We remove previously imported data via dev migrations and import the real data into regular migrations (bestuurseenheden, organen, organen in time and mandates).
The minister positions remain dev data because the codelist will change in a near future.

To test is, can you run these migrations against the Eredienst mandatenbeheer and Bedienarenbeheer to see if you have everything you expected ?

While doing so, I noticed one thing that will probably need to be fixed in the frontend : in Eredienst mandatenbeheer, if we don't select a time period, when creating a new position we get duplicated results in the dropdown list of Mandaat. Maybe the `endDate` query param should be set by default to the current bestuurorgaan end date ?

Thanks !


cecemel:
@claire-lovisa thanks looks like it is doing its thing. @Poltergeistz @Windvis can you try out? and check the little quirks claire mentioned?

Windvis:
@claire-lovisa Could you document the steps that are needed to reproduce the bug? I can't seem to reproduce it, but I probably misunderstood something 😄.

claire-lovisa:
@Windvis  Sure ! So I log in with the mock login account of `Bestuur van de eredienst Anglicaanse Kerk Sint-Boniface van Antwerpen`. From the main page I go to `Eredienst mandatenbeheer`, then `Voeg mandataris toe`, select a person. When opening the `Mandaat` dropdown, we have twice each position title.
 Instead, if, when going to `Eredienst mandatenbeheer` you select a bestuursperiode, we have each position title only once.

Capture d’écran du 2022-09-06 11-21-35

Poltergeistz:
LGTM for `Bedienarenbeheer`

claire-lovisa:
We discovered a bug in the classifications of bestuursorganen in OP, I'll fix it asap and update the files that are impacted in this PR

claire-lovisa:
Update: the issue is coming from a later version than the prod one so the data here is correct ! No change required on the import :)

Poltergeistz (approved):

PR #217, opened 2022-09-02
Climate 2 subsidy adjustments

FangioD:
Added easy way to differentiate between LEKP1.0 and LEKP2.0
Use in combination with PR with same name on frontend-loket


cecemel:
closing for #218

PR #224, opened 2022-09-06
Feature/prepare worship export

aatauil:
- Add new prepare-submissions-for-export service.
- restructure folder for public & worship config 
- add delta rule
- add migration for new PUBLICATION_CONCEPT

To be tested with updated prepare-submissions-for-export service:
https://github.com/lblod/prepare-submissions-for-export-service/tree/feature/worship-database

Note:
- delta currently matching in //anything. Its possible to make it only match on concept "send" and file "downloaded" but the prepare-submissions-service has a export.json file with different types eg:
```
    {
      "type": "http://mu.semte.ch/vocabularies/ext/SubmissionDocument",
      "pathToSubmission": "?submission <http://purl.org/dc/terms/subject> ?subject; \n a <http://rdf.myexperiment.org/ontologies/base/Submission>."
    },
```
So kinda afraid we will somehow miss submissions because we match only on those specific deltas.

- besluitenlijstOptions might be incomplete (seems short) could the person reviewing double check this? 
- regulationTypes not needed?


config/delta-producer/submissions/prepare/worship/config.json
```diff
@@ -0,0 +1,34 @@
+{
+  "decisionTypes": [
+    "https://data.vlaanderen.be/id/concept/BesluitDocumentType/8e791b27-7600-4577-b24e-c7c29e0eb773",
+    "https://data.vlaanderen.be/id/concept/BesluitType/b25faa84-3ab5-47ae-98c0-1b389c77b827",
```
claire-lovisa, line 4: I don't see this one in the GSheet, where did you find it ?

config/delta-producer/submissions/prepare/worship/config.json
```diff
@@ -0,0 +1,34 @@
...
+    "https://data.vlaanderen.be/id/concept/BesluitType/e44c535d-4339-4d15-bdbf-d4be6046de2c",
+    "https://data.vlaanderen.be/id/concept/BesluitDocumentType/672bf096-dccd-40af-ab60-bd7de15cc461",
+    "https://data.vlaanderen.be/id/concept/BesluitType/79414af4-4f57-4ca3-aaa4-f8f1e015e71c",
+    "https://data.vlaanderen.be/id/concept/BesluitType/54b61cbd-349f-41c4-9c8a-7e8e67d08347",
+    "https://data.vlaanderen.be/id/concept/BesluitType/40831a2c-771d-4b41-9720-0399998f1873",
+    "https://data.vlaanderen.be/id/concept/BesluitDocumentType/18833df2-8c9e-4edd-87fd-b5c252337349",
+    "https://data.vlaanderen.be/id/concept/BesluitType/df261490-cc74-4f80-b783-41c35e720b46",
+    "https://data.vlaanderen.be/id/concept/BesluitType/f56c645d-b8e1-4066-813d-e213f5bc529f",
+    "https://data.vlaanderen.be/id/concept/BesluitDocumentType/2c9ada23-1229-4c7e-a53e-acddc9014e4e",
+    "https://data.vlaanderen.be/id/concept/BesluitType/3fcf7dba-2e5b-4955-a489-6dd8285c013b"
```
claire-lovisa, line 14: This one can't be created in Loket currently because the sender doesn't have an account in Loket, should we still keep it ?

config/delta-producer/worship-submissions/prepare/config.json
```diff
@@ -0,0 +1,33 @@
...
+    "https://data.vlaanderen.be/id/concept/BesluitType/79414af4-4f57-4ca3-aaa4-f8f1e015e71c",
+    "https://data.vlaanderen.be/id/concept/BesluitType/54b61cbd-349f-41c4-9c8a-7e8e67d08347",
+    "https://data.vlaanderen.be/id/concept/BesluitType/40831a2c-771d-4b41-9720-0399998f1873",
+    "https://data.vlaanderen.be/id/concept/BesluitDocumentType/18833df2-8c9e-4edd-87fd-b5c252337349",
+    "https://data.vlaanderen.be/id/concept/BesluitType/df261490-cc74-4f80-b783-41c35e720b46",
+    "https://data.vlaanderen.be/id/concept/BesluitType/f56c645d-b8e1-4066-813d-e213f5bc529f",
+    "https://data.vlaanderen.be/id/concept/BesluitDocumentType/2c9ada23-1229-4c7e-a53e-acddc9014e4e",
+    "https://data.vlaanderen.be/id/concept/BesluitType/3fcf7dba-2e5b-4955-a489-6dd8285c013b" 
+  ],
+  "besluitenlijstOptions": [
```
cecemel, line 15: I don't find besluitenlijst in https://docs.google.com/spreadsheets/d/1NnZHqaFnNToE-aZMiyDI1QIPhHP5EG8i39KGFhTazlg/edit#gid=0

cecemel:
@aatauil I am also a little bit concerned about performance; so any clues would be nice. Could you explain your concern? (here or in PM)

cecemel:
@claire-lovisa @aatauil I think we need more specific filters though. Take e.g notulen. The ones from Erediensten is conficential, from other bestuurseenheden not. Now all notulen from all erediensten will also end up in the public decisions

claire-lovisa:
@cecemel Indeed. Didn't look how the code works in depth, but isn't the flag made for this ? If the flag is different from the public decisions db, it shouldn't be exported there ?

aatauil:
@cecemel I was actually hoping you could explain :p So you suggested to only add deltas for download & send status. But in the export.json you have config helping you match a submission based off indirect info. e.g. 
```
{
      "type": "http://mu.semte.ch/vocabularies/ext/TaxType",
      "pathToSubmission": "?submission <http://www.w3.org/ns/prov#generated> ?formData; \n a <http://rdf.myexperiment.org/ontologies/base/Submission>. \n ?formData <http://mu.semte.ch/vocabularies/ext/taxType> ?subject."
},
``` 
If we would add the 2 delta match rules the contents of export.json should change to the following types:  "Concept" for send status match and "FileDataObject" for download status (I think). But a lot of redundency disappears which makes me think we might miss a submission

aatauil:
> @cecemel Indeed. Didn't look how the code works in depth, but isn't the flag made for this ? If the flag is different from the public decisions db, it shouldn't be exported there ?

yes indeed, they are independently flagged.

cecemel:
@aatauil @c

> @cecemel Indeed. Didn't look how the code works in depth, but isn't the flag made for this ? If the flag is different from the public decisions db, it shouldn't be exported there ?

@aatauil @claire-lovisa it is rather to make sure that two instances of the service can make the difference.

claire-lovisa:
So as discussed, the PR is extended by using only one instance of the prepare-submissions service that flags submissions based on defined rules.
I re-did by accident some work that Achraf already did in #231 and noticed it too late, so I stole two commits from there. This PR is now about the prepare-submission AND the producer config. It makes it less convenient to review, sorry about that

PR #248, opened 2022-09-27
Change file model `nfo:fileCreated` to `dct:created`

benjay10:
Change all references of `nfo:fileCreated` to `dct:created`. Due to a small documentation confusion this property has not been updated correctly at the time.

This PR should include a SPARQL query that I tested on a copy on some database with test data. There shouldn't be a single triple with the `nfo:fileCreated` as a property, even if the update query was careful enough to only change instances about files.

I could not find any other reference to `nfo:fileCreated` in any other file or any other used services for this stack.


cecemel:
@benjay10 
on prod; i find 
```
PREFIX nfo: <http://www.semanticdesktop.org/ontologies/2007/03/22/nfo#>
SELECT COUNT(DISTINCT( ?s)) ?what{
  ?s nfo:fileCreated ?created; a ?what.
}
group by ?what
```
a lot of results

The modules responsible for interacting with files from frontend
- inzendingen voor toezicht (ember-submission-fom-fields addon)
- subsidies  (ember-submission-fom-fields addon)
- bbcdr (the file component)
-  berichtencentrum t(he file component)

Services for fetching files to ingest:
- berichtencentrum sync with kalliope

Other peculiarity:
- when you do the migration: exclude all publication graphs (see docker-compose)

benjay10:
All the ember related things access the data via the JSON:API that is translated by mu-cl-resources, so if mu-cl-resources parses and writes the correct property, the JSON:API should remain completely unchanged. Or am I still missing something?
I put it to the test and I can see and download the files in the 4 locations you point out. I had already tested that before. (The downloads fail for me because I don't have physical files related to the test data. The error message from the file service does indicate to me that it attempts to download everything correctly.)

The ember-submission-form-fields addons just deals with creating forms? Is the proposed change in config, or something similar then not enough to adopt the new property? Anyway, I can't find a direct reference of `nfo:fileCreated` at all, so there should be no need to change anything.

The berichtencentrum-sync-with-kalliope-service just synchronises messages? Then there is also nothing to change, except for where the messages are created in the first place. This other stack also needs to adopt the correct model for files.

Your query might also to bring up other types of data that are not files, and I don't think we should touch those things as they might still rely on that property for their respective data model, except for all the `nfo:RemoteDataObject`, `nfo:LocalFileDataObject` and `nfo:FileDataObject`. 

I will check the graphs to make the query more specific, but other than that, I don't know what else to do here.

cecemel:
@benjay10 ok thanks, i went over all other services and it seems no changes are needed. (sorry being extra cautious here)
Also rebased development cause already ol pr

PR #252, opened 2022-09-28
Clear out organen before 2023

claire-lovisa:
I noticed on prod that we already have worship organen in time for the periods before 2023, which is not supposed to be the case. Not sure how they got here, but it might conflict with a future data import.

We are going to import those organen back later on but with different start and end dates, so I suggest we clear them out to start from a clean sheet.

### DIFF
--- a/config/migrations/2022/20220829135928-erediesten-data/20220928092536-clear-erediensten-governing-bodies-before-2023.sparql
+++ b/config/migrations/2022/20220829135928-erediesten-data/20220928092536-clear-erediensten-governing-bodies-before-2023.sparql
@@ -1,0 +1,49 @@
+PREFIX mandaat: <http://data.vlaanderen.be/ns/mandaat#>
+PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
+PREFIX besluit: <http://data.vlaanderen.be/ns/besluit#>
+PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
+PREFIX org: <http://www.w3.org/ns/org#>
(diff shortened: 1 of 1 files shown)


cecemel (commented):
This needs to run before  https://github.com/lblod/app-digitaal-loket/pull/253/files
right?

config/migrations/2022/20220829135928-erediesten-data/20220928092536-clear-erediensten-governing-bodies-before-2023.sparql
```diff
@@ -0,0 +1,49 @@
...
+    ?bestuurseenheid a ?type .
+  
+    ?bestuursorgaan besluit:bestuurt ?bestuurseenheid .
+  
+    ?bestuursorgaanInTime mandaat:isTijdspecialisatieVan ?bestuursorgaan ;
+      mandaat:bindingStart ?start ;
+      ?pbestuursorgaanInTime ?obestuursorgaanInTime .
+
+    # We just keep the ones we created for time periods 2023-2026 and 2026-2029
+    FILTER( ?start < "2023-04-01T00:00:00"^^<http://www.w3.org/2001/XMLSchema#dateTime> )
```
Poltergeistz, line 34: According to #253, Organen dates are not `dateTime` anymore right ?

claire-lovisa: Indeed but this one will run before #252 (see timestamps) so when it'll run, it'll still be dateTimes and not dates yet !

claire-lovisa:
@cecemel @Poltergeistz Can I merge it ?

PR #257, opened 2022-10-05
lpdc tackle republish problem delta

jsz4n:


benjay10 (approved):
These changes make sense!

PR #260, opened 2022-10-06
Add worship organen and mandates before 2023

claire-lovisa:


Poltergeistz (approved):

cecemel:
@Poltergeistz did you check the behavior of the bestuursperiode-selector in the frontend?
Also, perhaps it is worth making more obvious for what start end  and end date they are selecting the mandates.
Perhaps in the select box add in the template: 
```
Secretaris ( {{mandate.blalbla..bestuursorgaan.start - mandate.blabla.bestuursorgaan.end }}) 
```
Note also then to take into account when mandates have more than bestuursorgaan linked (probs the little helper @Windvis wrote could help)
-> make a ticket in jira if you start implementing this feature

PR #274, opened 2022-10-20
lpdc add mandatory executing authority

jsz4n:
For DL-4390

### DIFF
--- a/config/lpdc-management/characteristics/form.ttl
+++ b/config/lpdc-management/characteristics/form.ttl
@@ -200,7 +200,6 @@ ext:authorityPg a form:PropertyGroup;
       sh:path lpdcExt:hasExecutingAuthority ;
       form:validations [
         a form:RequiredConstraint ;
-        form:language "en";
         form:grouping form:Bag ;
         sh:resultMessage "Dit veld is verplicht." ;
         sh:path lpdcExt:hasExecutingAuthority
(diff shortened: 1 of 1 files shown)


config/lpdc-management/characteristics/form.ttl
```diff
@@ -198,6 +198,13 @@ ext:authorityPg a form:PropertyGroup;
       form:help "Meerdere overheden kunnen worden toegewezen";
       sh:order 30 ;
       sh:path lpdcExt:hasExecutingAuthority ;
+      form:validations [
+        a form:RequiredConstraint ;
+        form:language "en";
```
cecemel, line 203: This is only needed when validating something that expects a lang string. 
In our case this is a code list entry.

cecemel: So remove this

config/lpdc-management/characteristics/form.ttl
```diff
@@ -198,6 +198,13 @@ ext:authorityPg a form:PropertyGroup;
...
+        form:grouping form:Bag ;
+        sh:resultMessage "Dit veld is verplicht." ;
+        sh:path dct:description
```
cecemel, line 206: the sh:path specifies on what predicate the subject should be validated again. It might look redundant, given the way it is specified now, because inline, but it should actually `sh:path lpdcExt:hasExecutingAuthority ` too.
Actually, the fact that it didn't work, should be very obvious with even light testing: Even if this field would have been filled in, you wouldn't be able to send it

cecemel (changes requested):

cecemel:
@jsz4n thanks

PR #283, opened 2022-11-04
chore remove old identifiers

Asergey91:
note: this might be a bit too aggressive, i tested and nothing straight up broke but might affect some parts i am not aware of


cecemel:
@Asergey91 sorry i am gonna keep it like it is.

PR #291, opened 2022-11-29
lpdc publish deleted services

jsz4n:
DL-4560


config/delta/lpdc-publish.js
```diff
@@ -20,5 +20,27 @@ export default [
       ignoreFromSelf: true,
       optOutMuScopeIds: [ "http://redpencil.data.gift/id/concept/muScope/deltas/initialSync" ]
     }
+  },
+  {
```
cecemel, line 24: mind we needed an extra rule, this eases postprocessing delta in lpdc-publish

PR #305, opened 2022-12-23
Feature/subsidy stadsvernieuwing

FangioD:


cecemel:
First step is ok. Now wee need to figure out how to resolve the part: limit the amount of bestuurseenheden who can see this.

PR #315, opened 2023-01-12
fix languages

Asergey91:
this fixes current exisitng languages but doesnt fix the ones that didnt get imported


cecemel (changes requested):
I think the last file should not have been committed

PR #320, opened 2023-01-16
Add a "review-status" relationship to the public services resource

Asergey91:


config/migrations/2023/20230116120500-add-status-concepts.ttl
```diff
@@ -0,0 +1,24 @@
...
+
+<http://lblod.data.gift/concepts/cf22e8d1-23c3-45da-89bc-00826eaf23c3> a <http://www.w3.org/2004/02/skos/core#Concept>;
+  <http://www.w3.org/2004/02/skos/core#inScheme> <http://lblod.data.gift/concept-schemes/effb1058-d469-4064-a7a6-9d98ba081962>;
+  <http://www.w3.org/2004/02/skos/core#prefLabel> "Deleted";
+  <http://www.w3.org/2004/02/skos/core#topConceptOf> <http://lblod.data.gift/concept-schemes/effb1058-d469-4064-a7a6-9d98ba081962>;
+  <http://mu.semte.ch/vocabularies/core/uuid> "cf22e8d1-23c3-45da-89bc-00826eaf23c3".
+
+<http://lblod.data.gift/concepts/176ec8a9-ffd6-4f0b-a302-84ef59b31c42> a <http://www.w3.org/2004/02/skos/core#Concept>;
+  <http://www.w3.org/2004/02/skos/core#inScheme> <http://lblod.data.gift/concept-schemes/effb1058-d469-4064-a7a6-9d98ba081962>;
+  <http://www.w3.org/2004/02/skos/core#prefLabel> "Reviewed";
```
Windvis, line 22: Thinking about it some more, do we actually need the "reviewed" status? Couldn't we simply interpret the absence of the triple as "no review required"?

That would also save us the work of having to set the "Reviewed" status on newly created public service instances.

cecemel: we could start with that yes

config/migrations/2023/20230116120500-add-status-concepts.ttl
```diff
@@ -0,0 +1,24 @@
+
+
+<http://lblod.data.gift/concept-schemes/effb1058-d469-4064-a7a6-9d98ba081962> a <http://www.w3.org/2004/02/skos/core#ConceptScheme>;
+  <http://www.w3.org/2004/02/skos/core#prefLabel> "Statuses of public services";
```
cecemel, line 4: It's not a status of the public service, it kinda summarizes what the state of the linked conceptual service has been.

`Statuses of linked conceptual public service`

config/resources/lpdc.json
```diff
@@ -68,6 +68,11 @@
           "predicate": "pera:language",
           "cardinality": "many"
         },
+        "review-status":{
+          "target": "concept",
+          "predicate": "ext:status",
```
cecemel, line 73: a more expressive status in name?

config/migrations/2023/20230116120500-add-status-concepts.ttl
cecemel: Second thought. We want to say two things.
We want to inform the user some changes happened on the concept and they might want to review their status.
We also want to say what happened to the status; updated or deleted.

config/resources/lpdc.json
```diff
@@ -68,6 +68,11 @@
           "predicate": "pera:language",
           "cardinality": "many"
         },
+        "review-status":{
+          "target": "concept",
+          "predicate": "ext:reviewStatus",
+          "cardinality": "one"
+        },
```
Windvis, line 75: I think this should best be moved to the "public-services" resource since it only applies to that one, right?

Windvis:
I think we also need to update the mu-auth config so we can update the status from the frontend (to mark things as reviewed). At the moment the changes aren't persisted.

PR #323, opened 2023-01-18
Adding bestuurseenheid read only flag

Poltergeistz:
DL-4722

~~Adding the migration for a bestuurseenheid (molokai)~~
Adding the config in mu-cl-resource

~~For now the flag is named `hasEditRight` because the front-end is configured that way but we can always change it if needed.
The hasEditRight flag will be set to true when bestuurseenheiden have access role eredienst-mandatenbeheer/eredienst-bedienarenbeheer~~


config/migrations/dev/2023/20230119110751-add-read-only-module-flag-to-bestuurseenheid.sparql
```diff
@@ -0,0 +1,19 @@
+# Adding viewOnly flag to the bestuurseenheid who has access to the eredienst modules
```
cecemel, line 1: don't commit this file, this will run in production. `mkdir config/dev-migrations/2023/` and add a README.md there that this won't run by default, and if you need these that you'll have to run them manually.

cecemel (changes requested):

config/resources/organisation.lisp
```diff
@@ -11,7 +11,8 @@
                 (:alternatieve-naam :string-set ,(s-prefix "skos:altLabel"))
                 (:wil-mail-ontvangen :boolean ,(s-prefix "ext:wilMailOntvangen")) ;;Voorkeur in berichtencentrum
                 (:mail-adres :string ,(s-prefix "ext:mailAdresVoorNotificaties"))
-                (:is-trial-user :boolean ,(s-prefix "ext:isTrailUser")))
+                (:is-trial-user :boolean ,(s-prefix "ext:isTrailUser"))
+                (:has-view-only-module :string-set ,(s-prefix "ext:hasViewOnlyModule")))
```
Windvis, line 15: nitpick: I would call this something like `view-only-modules` since it contains the actual list of module names.

Windvis (approved):
Works as expected 👍.

PR #333, opened 2023-01-26
[Toezicht Module] - APB - Fix form RPR

Poltergeistz:
DL-4748

Fixes add link/upload file fields for administrative unit APB


Windvis (approved):
I tested it before the changes to verify the bug, and using the branch made the form work as expected 👍.

PR #337, opened 2023-02-01
[Toezicht module] Delete 53 concepts from automatic submission

Poltergeistz:
DL-4482

Adding the migration file for deleting the concepts from the automatic submission.

### DIFF
--- a/config/migrations/2023/20230201113240-remove-automatic-submissions-gemeente-bilzen.sparql
+++ b/config/migrations/2023/20230201113240-remove-automatic-submissions-gemeente-bilzen.sparql
@@ -1,0 +1,27 @@
+# There is 53 automatic submissions concepts to delete from gemeente Bilzen in toezicht module.
+
+PREFIX meb: <http://rdf.myexperiment.org/ontologies/base/>
+PREFIX dct: <http://purl.org/dc/terms/>
+PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
+PREFIX pav: <http://purl.org/pav/>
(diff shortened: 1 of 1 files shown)


aatauil (changes requested):
Scope of this migration is too broad. This will delete all submission concepts event the ones that they dont want deleted. In the screenshot only 13 submissions need removing. We need to somehow filter the fault submissions. Filtering by date also seems a bit risky unless we have the exact range but just saying "after 25 november" might still risk removing submissions that have come in in the meanwhile. 

Perhaps a sane thing is to filter on submissions without a "type dossier" as thats an anomaly in my eyes.

@cecemel might need input on this

Poltergeistz:
> Scope of this migration is too broad. This will delete all submission concepts event the ones that they dont want deleted. In the screenshot only 13 submissions need removing. We need to somehow filter the fault submissions. Filtering by date also seems a bit risky unless we have the exact range but just saying "after 25 november" might still risk removing submissions that have come in in the meanwhile.
> 
> Perhaps a sane thing is to filter on submissions without a "type dossier" as thats an anomaly in my eyes.
> 
> @cecemel might need input on this

Thanks for the review :+1: . Indeed as type dossier is not provided we might use that but I don't know if there is submissions that share the same 'empty' type dossier.

aatauil:
Perhaps a combination all empty fields rather then just type dossier. Perhaps  type dossier AND orgaan, I think orgaan should always be there. Could you check if there exists submissions without orgaan?

PR #339, opened 2023-02-03
Add missing subsidy for Deinze Municipality

Riadabd:
(Addresses DL-4814) Add the "stadsvernieuwing" subsidy to "Gemeente Deinze" as it was missing from the previous subsidy migration.

### DIFF
--- a/config/migrations/2023/20230203114112-add-missing-stadsvernieuwing-subsidy-for-Deinze.sparql
+++ b/config/migrations/2023/20230203114112-add-missing-stadsvernieuwing-subsidy-for-Deinze.sparql
@@ -1,0 +1,35 @@
+PREFIX mobiliteit: <https://data.vlaanderen.be/ns/mobiliteit#>
+PREFIX lblodSubsidie: <http://lblod.data.gift/vocabularies/subsidie/>
+PREFIX dct: <http://purl.org/dc/terms/>
+PREFIX subsidie: <http://data.vlaanderen.be/ns/subsidie#>
+PREFIX gleif: <https://www.gleif.org/ontology/Base/>
+PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
+PREFIX cpsv: <http://purl.org/vocab/cpsv#>
+PREFIX common: <http://www.w3.org/2007/uwa/context/common.owl#>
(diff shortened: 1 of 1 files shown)


config/migrations/2023/20230203114112-add-missing-stadsvernieuwing-subsidy-for-Deinze.sparql
```diff
@@ -0,0 +1,35 @@
...
+  }
+}
+
+WHERE {
+  BIND(<http://data.lblod.info/id/bestuurseenheden/d93451bf-e89a-4528-80f3-f0a1c19361a8> as ?bestuureenheid) # Gemeente Deinze
+  
+  ?bestuureenheid mu:uuid ?uuid .
+  BIND(IRI(CONCAT("http://mu.semte.ch/graphs/organizations/", STR(?uuid), "/LoketLB-subsidies")) as ?newGraph)
+
+  BIND(<http://lblod.data.info/id/subsidy-measure-offers/c10fc76f-effe-430e-822f-60f03f575302> as ?SubsidiemaatregelAanbod)
```
Poltergeistz, line 27: I think this should be `<http://data.lblod.info/id/subsidy-measure-offers/c10fc76f-effe-430e-822f-60f03f575302` instead of `lblod.data.info`. Same for the line below.

Riadabd: Nice catch! It seems though a lot of the previous subsidy migrations have `lblod.data.info` instead of `data.lblod.info` in the URIs, and I don't know if it's a typo or by design.

Poltergeistz: To be honest, I don't know what's the design of the url :thinking: . From what I see, it looks like `lblod.data` is formated that way for `.gift` but I'm not sure for `.info`

cecemel: so great typo catching, let's keep things as is for now, we'll have to clean up later

PR #355, opened 2023-02-28
OP Public Consumer

bdevloed:
wip, needs testing and cleanup


cecemel:
@bdevloed (see also: https://gist.github.com/cecemel/2443bf343e4b94b23e835d6869fc6611#24-data-update-consumer-from-op-data-open)
I think it be safer to be less agressive on the delete:
- working with a whitelist of predicates
-  also; I am not sure i would delete in ALL graphs; need to further check where the data resides

config/reasoner/op2dl/main/mapping.n3
```diff
@@ -0,0 +1,161 @@
...
+
+# DL and OP both use persoon:gebruikteVoornaam but with different semantics (!)
+# Hence the mapping is in two steps.
+# The first step maps the DL persoon:gebruikteVoornaam to the OP counterpart
+# and drops the DL persoon:gebruikteVoornaam statements with e:becomes
+# A second reasoning will map DL foaf:name to OP persoon:gebruikteVoornaam
+# Doing both at once would lead all DL foaf:names to incorrectly become foaf:givennames
+# in OP
+
+# First name
```
cecemel, line 30: I guess we don't expect persons right?

config/reasoner/op2dl/main/mapping.n3
```diff
@@ -0,0 +1,161 @@
...
+    ?s persoon:gebruikteVoornaam ?o
+  } e:becomes {
+    ?s foaf:name ?o
+  }
+} => {
+  ?s foaf:name ?o
+}.
+
+
+{
```
cecemel, line 56: nor mandatarissen right?

config/reasoner/op2dl/main/persoon_https_to_http.n3
```diff
@@ -0,0 +1,534 @@
+PREFIX : <#>
```
cecemel, line 1: didn't expect this one?

config/reasoner/op2dl/main/rrn.n3
```diff
@@ -0,0 +1,45 @@
+PREFIX : <#>
```
cecemel, line 1: neither this one :-)

config/reasoner/op2dl/main/mapping.n3
bdevloed: Idd, it's just a generic mapping (I inverted the OP -> DL mapping). There's no performance drawback and keeps the reasoner config generic and re-usable

config/reasoner/op2dl/main/mapping.n3
bdevloed: cf above

config/reasoner/op2dl/main/persoon_https_to_http.n3
bdevloed: It's a relative uri. I used this namespace for the backward arrow rules (i.e. date -> dateTime conversions) which are defined inside the mapping file.

config/reasoner/op2dl/main/rrn.n3
bdevloed: cf above

bdevloed:
To Do: Update delta-consumer image tag once context feature is released.

cecemel (commented):
Thanks!
Still want to test it though. 
We'll have to revise the comment about the mu-semte.ch thing, but not now.
The intermediate graphs, where are these stored now?

cecemel:
-> I tested it, it's quite spectacular a system inflating itself :-)

config/op-consumer/public/delta-context-config.js
```diff
@@ -0,0 +1,193 @@
...
+              adms:identifier ?identifier.
+          }
+        }`
+    },
+    {
+      trigger: { // subjectType or predicateValue
+        predicateValue: "generiek:lokaleIdentificator"
+      },
+      queryTemplate: (subject) => `
+        ${PREFIXES}
```
cecemel, line 72: Still have to wrap my mind about it, but it blurs a bit the line between what the reasoner does, and what the context does. Implicitly, here, the context decides what extra triples can be deleted/inserted, whereas it feels more here it's a reasoner to do it.

cecemel: And i am not sure  EYE N3 Reasoner  supports this kind of expressivity

config/op-consumer/public/delta-sync-dispatching.js
```diff
@@ -0,0 +1,116 @@
...
+
+      const transformedDeletes = await transformStatements(fetch, deletesWithContext);
+
+      if (!transformedDeletes.length) {
+        console.log(`Warn: Delete statements mapped to empty result.`);
+        console.log(`Input: ${deletesWithContext}`);
+        console.log(`Output: ${transformedDeletes}`);
+      } else {
+        await deleteFromTargetGraph(lib, transformedDeletes);
+        await insertIntoDebugGraph(lib, transformedDeletes);
```
cecemel, line 93: wondering if it still has a good value/cost ratio given all intermediate buffers we have

config/op-consumer/public/delta-context-config.js
bdevloed: The query here is only executed to (try) to add context to the deltas if the trigger matches.
The reasoner later does the mapping and filtering based on the configured rules and queries.

The reasoner does this in this scenario with deltas+context. Or just with the whole ingest graph during the initial sync (where this config is irrelevant).

So, not sure what you mean with the expressivity. If it's fetching additional data as input at runtime, then indeed it doesn't. Even though there technically is a built-in `log:semantics` which can fetch data - I think it's probably not most suitable option.

Let me know though if I misinterpreted your comment.

config/op-consumer/public/delta-sync-dispatching.js
bdevloed: Specifically the debug graph, right? Ok for me to drop it.

data/db/toLoad/output000001.nq.gz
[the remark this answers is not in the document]
bdevloed: should be removed - this is by acccident

config/op-consumer/public/util.js
```diff
@@ -0,0 +1,229 @@
...
+    MAX_DB_RETRY_ATTEMPTS,
+    SLEEP_BETWEEN_BATCHES,
+    SLEEP_TIME_AFTER_FAILED_DB_OPERATION,
+    'INSERT');
+}
+
+async function transformIngestGraph(fetch) {
+  console.log(`Transforming ingest graph: ${INGEST_GRAPH}`);
+
+  const response = await fetch(`http://reasoner/reason/op2dl/main?data=${encodeURIComponent(`${INGEST_DATABASE_ENDPOINT}?default-graph-uri=&query=CONSTRUCT+%7B%0D%0A%3Fs+%3Fp+%3Fo%0D%0A%7D+WHERE+%7B%0D%0A+GRAPH+<${INGEST_GRAPH}>+%7B%0D%0A%3Fs+%3Fp+%3Fo%0D%0A%7D%0D%0A%7D&should-sponge=&format=text%2Fplain&timeout=0&run=+Run+Query`)}`);
```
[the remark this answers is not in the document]
bdevloed: I agree this isn't pretty. And happy to change this for an alternative, given it's able to handle bulk transactions within a reasonable time - this function is only called during the initial sync. Earlier attempts to do the initial sync compliant to the stack took a job from minutes to days - even without the querying for additional context. And added a lot of extra complexity to handle the arbitrary partitioning of sets of triples. Hence I saw this as the least bad option.
However, happy to discuss and improve this if possible.

PR #362, opened 2023-03-10
Added migration to flush for OP - sync v1

cecemel:
@bdevloed this migration flushes the data that will be managed in the first version of OP sync. It targets `feature/op-public-consumer`. You can just merge.


claire-lovisa (commented):
I'll let @bdevloed give the final approve, but I went over it carefully and I think flushes what needs to be flushed :+1:

PR #364, opened 2023-03-13
Remove redundant person and mandataris URIs for raadslid Rotselaar

Riadabd:
(Addresses DL-4715) Remove redundant person URIs for raadslid Rotselaar, in addition to the mandataris URIs linked to those person URIs.


cecemel:
@Riadabd I did some changes to the migration, these duplicates are usually tricky.

PR #366, opened 2023-03-16
Add service that creates mock accounts automatically

claire-lovisa:
This PR adds a service that automatically creates missing mock-login accounts (and removed outdated ones). It becomes handy now that admin units arrive to the application via a delta sync, and is especially useful for the control environment.

About the technical choices made in the service:
1. I chose a cron job trigger type and not a delta trigger type because the cron job way is lighter on the stack and to implement, and mock accounts are not so urgent to get so we don't need it right away when the bestuur arrived in the stack
2. A config file has to be added in the application using the service. That way, the service can be reused over multiple stacks working with different roles.

The service can be found here https://github.com/lblod/update-bestuurseenheid-mock-login-service

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -437,6 +437,14 @@ services:
       - "logging=true"
     restart: always
     logging: *default-logging
+  update-bestuurseenheid-mock-login:
+    image: lblod/update-bestuurseenheid-mock-login-service:0.1.0
+    volumes:
+      - ./config/mock-login:/config
+    labels:
+      - "logging=true"
+    restart: always
+    logging: *default-logging
   ################################################################################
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -437,6 +437,14 @@ services:
       - "logging=true"
     restart: always
     logging: *default-logging
+  update-bestuurseenheid-mock-login:
+    image: lblod/update-bestuurseenheid-mock-login-service:0.1.0
+    volumes:
+      - ./config/mock-login:/config
```
nbittich, line 443: is this required?

nbittich (approved):

PR #371, opened 2023-03-22
[DL-4974] Fix vendor access management for worship services

Windvis:
This allows the user to manage vendors for worship related administrative units. I'm not sure if this is the correct fix but it seems the user also had all other roles already.

It's possible that this should be hotfixed.

### DIFF
--- a/config/dispatcher/dispatcher.ex
+++ b/config/dispatcher/dispatcher.ex
@@ -642,7 +642,7 @@ defmodule Dispatcher do
   end
 
   match "/worship-services/*path" do
-    forward conn, path, "http://resource/worship-services/"
+    forward conn, path, "http://cache/worship-services/"
   end
 
   match "/recognized-worship-types/*path" do
--- a/config/authorization/config.ex
+++ b/config/authorization/config.ex
@@ -347,7 +347,18 @@ defmodule Acl.UserGroups.Config do
                     constraint: %ResourceConstraint{
                       resource_types: [
                         "http://mu.semte.ch/vocabularies/ext/Vendor"
-                      ] } } ] },
+                      ] } },
+                  %GraphSpec{
+                    graph: "http://mu.semte.ch/graphs/public",
+                    constraint: %ResourceConstraint{
+                      resource_types: [
+                        "http://data.vlaanderen.be/ns/besluit#Bestuurseenheid",
+                      ],
+                  predicates: %NoPredicates{
+                    except: [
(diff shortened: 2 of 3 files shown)


Windvis:
Switching to mu-cache did indeed fix the caching issue.

config/migrations/2023/20230322170100-fix-vendor-management-mock-user-roles.sparql
```diff
@@ -0,0 +1,13 @@
+PREFIX ext: <http://mu.semte.ch/vocabularies/ext/>
```
cecemel, line 1: @Windvis cool!  I am not sure this migration is needed. :-)

 Yeah totally forgot that we modify bestuurseenheden that are in public and read only.
BUT I THINK you can use similar construct https://github.com/lblod/app-digitaal-loket/blob/master/config/authorization/config.ex#L326
for this https://github.com/lblod/app-digitaal-loket/blob/master/config/authorization/config.ex#L340 for the public graph and that specific predicate

config/dispatcher/dispatcher.ex
```diff
@@ -642,7 +642,7 @@ defmodule Dispatcher do
   end
 
   match "/worship-services/*path" do
-    forward conn, path, "http://resource/worship-services/"
+    forward conn, path, "http://cache/worship-services/"
```
cecemel, line 645: literally no idea why it was in resource in the first place. so :+1:

cecemel (changes requested):

config/migrations/2023/20230322170100-fix-vendor-management-mock-user-roles.sparql
Windvis: I added them because that user seemed to have all other roles as well. I can take a look at the mu-auth config but that thing is still an enigma to me at the moment.

config/dispatcher/dispatcher.ex
Windvis: [This comment](https://github.com/lblod/app-digitaal-loket/pull/371#issuecomment-1479898750) explains it a bit. It was copied that way from OP but I don't think it's needed for loket.

config/migrations/2023/20230322170100-fix-vendor-management-mock-user-roles.sparql
cecemel: i will do it

PR #374, opened 2023-03-24
Update status and add orgaan to orgaan

claire-lovisa:
OP-2231 & DL-4980

@cecemel it will fix the submissions that can't be sent because of the missing orgaan :)


claire-lovisa:
@cecemel probably fixed indeed but I can't double check as I'm on holidays this week, could some else verify ? (@bdevloed @nbittich ) thanks :)

claire-lovisa:
@cecemel I checked, data has been flowing properly to Loket :) Closing this PR !

PR #376, opened 2023-03-29
Allow the executing authority field to be optional in some cases

Windvis:
It doesn't work as intended yet. The field is hidden until a value is selected for the "executing authority level".

It requires this change to be merged and released: https://github.com/lblod/submission-form-helpers/pull/42

### DIFF
--- a/config/lpdc-management/characteristics/form.ttl
+++ b/config/lpdc-management/characteristics/form.ttl
@@ -4,6 +4,7 @@
 @prefix mu: <http://mu.semte.ch/vocabularies/core/> .
 @prefix displayTypes: <http://lblod.data.gift/display-types/> .
 @prefix ext: <http://mu.semte.ch/vocabularies/ext/> .
+@prefix dvc: <https://productencatalogus.data.vlaanderen.be/id/concept/UitvoerendBestuursniveau/> .
 
 @prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>.
 @prefix skos:	<http://www.w3.org/2004/02/skos/core#> .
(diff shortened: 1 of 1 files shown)


config/lpdc-management/characteristics/form.ttl
```diff
@@ -191,8 +193,52 @@ ext:authorityPg a form:PropertyGroup;
...
+          # Tip: you can use form:valuesNotIn to exclude from a list (if triples are found matching the path)
+          form:valuesNotIn (
+            <https://productencatalogus.data.vlaanderen.be/id/concept/UitvoerendBestuursniveau/Derden>
+          );
+        ] ;
+      form:hasFieldGroup ext:temp-fg-a-result .
+  
+  ext:temp-b a form:ConditionalFieldGroup ;
+      form:conditions
+      [ a form:SingleCodelistValue ;
```
Windvis, line 224: SingleCodeListValue doesn't seem to work because we use the "conceptSelector" field so the code list values aren't stored in the meta graph.

config/lpdc-management/characteristics/form.ttl
```diff
@@ -1,4 +1,5 @@
 @prefix form: <http://lblod.data.gift/vocabularies/forms/> .
+@prefix fieldGroups: <http://data.lblod.info/field-groups/> .
```
Windvis, line 2: Not needed I think. i copied it from a v1 form but it doesn't seem to be used here.

config/lpdc-management/characteristics/form.ttl
Windvis: Maybe this one works? https://github.com/lblod/submission-form-helpers/blob/master/src/constraints/exact-value.js

config/lpdc-management/characteristics/form.ttl
```diff
@@ -191,8 +194,46 @@ ext:authorityPg a form:PropertyGroup;
       form:displayType displayTypes:conceptSelector ;
       sh:group ext:authorityPg .
 
+
+  ##########################################################
+  # Conditionally show the required or not-required variant of the uitvoerende overheid field 
+  ##########################################################
+  ext:temp-fg-a a form:FieldGroup .
```
Windvis, line 201: the temp uris still need to be replaced

Windvis:
We went for the completely optional field for now.

PR #393, opened 2023-04-17
Make path fields unique

Riadabd:
(Addresses DL-5025) Ensure that all fields have unique `sh:path` values so that they don't override each other when the form is saved and reloaded.

To test:
* Open the `Stadsvernieuwing - projectsubsidie` through `Gemeente Deinze` (or any gemeente that has this subsidy)
* Submit the first `Aanvraag` form
*  Add text in the different large text input fields in the `Aanvraag (2e fase)` form
* Save the form and reload it
* No text field should override the other

### DIFF
--- a/config/subsidy-application-management/forms/urban-renewal/application-phase2/versions/20221218090000/form.ttl
+++ b/config/subsidy-application-management/forms/urban-renewal/application-phase2/versions/20221218090000/form.ttl
@@ -313,18 +313,18 @@ ext:samenvattingPropertyGroup
   ext:summaryField
       a                 form:Field ;
       sh:order          1 ;
-      sh:path           dct:description ;
+      sh:path           ( ext:samenvattingSummaryDescription dct:description ) ;
       form:validations
                         [ a                 form:MaxLength ;
                           form:grouping     form:MatchEvery ;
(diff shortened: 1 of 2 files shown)


config/subsidy-application-management/forms/urban-renewal/application-phase2/versions/20221218090000/form.ttl
```diff
@@ -313,18 +313,18 @@ ext:samenvattingPropertyGroup
...
       sh:order          1 ;
-      sh:path           dct:description ;
+      sh:path           ( ext:samenvattingSummaryDescription dct:description ) ;
       form:validations
                         [ a                 form:MaxLength ;
                           form:grouping     form:MatchEvery ;
                           form:max          "700" ;
                           sh:resultMessage  "Max. karakters overschreden." ;
-                          sh:path            dct:description
+                          sh:path           ( ext:samenvattingSummaryDescription dct:description )
```
cecemel, line 322: @Riadabd it is not clear why you include that extra paths. why not just 'ext:samenvattingSummaryDescription'

cecemel (changes requested):

Windvis (approved):

PR #400, opened 2023-04-25
Remove English requirement when YourEurope is not chosen

Riadabd:
(Addresses DL-3983) Removes the English translation requirements when `YourEurope` is not chosen as the publication channel in the `Eigenschappen` tab in a LPDC form.

This is to be tested in tandem with the following PR for `lpdc-management-service`: https://github.com/lblod/lpdc-management-service/pull/32.

To test:
* Log in as any Gemeente
* Go to the `Producten- en dienstencatalogus` module
* Add an existing public service or create a completely new one
* Toggle the publication channel fields in the `Eigenschappen` tab between empty and `YourEurope`
* The sections listed below should switch from `Verplicht` (when `YourEurope` is chosen) to nothing (when `YourEurope` is not chosen)

If it is chosen, `Titel`, `Beschrijving`, `Voorwaarden`, `Bewijsstukken` (Nested inside `Voorwaarden`), `Procedure`, `Website procedure` (Nested inside `Procedure`), `Kosten`, `Financiële Voordelen`, `Meer info` sections should have `Verplicht` alongside the English title and description fields.


config/lpdc-management/content/form-without-english-requirement.ttl
Windvis, line 1: Instead of duplicating the forms (and having to keep them in sync when we do changes) we had decided to use a different solution:

- remove the "required" constraints from all `@en` fields
- move them into a new english-constraints.ttl file (or whatever name makes sense)
- for the title field that would look like this:
```turtle
#add prefixes here so it's still a valid turtle file (in case we ever add linters)

  extEng:titleF 
      form:validations [
        a form:RequiredConstraint ;
        form:language "en";
        form:grouping form:Bag ;
        sh:resultMessage "Dit veld is verplicht." ;
        sh:path dct:title
      ].
```

- In the micro service, we can then simply read this file if needed, and concatenate it with the regular form.ttl string before sending it to the client.

The end result _should_ be the same, but we won't have to maintain multiple forms that are almost duplicates of each other.

Windvis (changes requested):

config/lpdc-management/content/form-without-english-requirement.ttl
Riadabd: It slipped my mind completely that I can append to the regular form in the code as if it's a regular string. Concerning the generators, would it be possible to insert them at run-time as well?

```
##########################################################
  #  Generator: Meer Info
  ##########################################################
  ext:websitesG a form:Generator;
    form:prototype [
      form:shape [
        a schema:WebSite ;
        dct:description " "@nl;
        dct:title " "@nl;
        schema:url "";
        dct:description " "@en;
        dct:title " "@en;
      ];
    ];
    form:dataGenerator form:addMuUuid.
```

At the moment, they look like this. Would it be possible to remove the `@en` lines and then append them for `form:shape` at runtime?

Windvis: I don't think so since it's a blank node and it would see it as 2 separate blank nodes I think? (not an expert so I might be wrong 😄). I would just ignore the generators for now. We can clean them up in a separate PR.

Windvis:
I released v0.22 of the microservice and bumped it here.

Could you add an entry to the changelog for this PR? I think it's good to go after that.

Windvis (approved):

PR #402, opened 2023-04-27
Allow changes to bestuurseenheid for vendor access management

benjay10:
DL-5012

Needed for https://github.com/lblod/frontend-vendor-access-management/pull/10.

Vendor access management needs to be able to set view-only-modules to the bestuurseenheid.


Windvis:
I think this may need to rebased on the latest development. Github is confused due to the recent merge issues.

benjay10:
> I think this may need to rebased on the latest development. Github is confused due to the recent merge issues.

I think with a merge it is back to a less confusing state.

PR #408, opened 2023-05-03
Convert `lpdc.json` to a `.lisp` file

Windvis:
`.lisp` is the prefered config format in this project.


config/resources/repository.lisp
Riadabd, line 1: I noticed `"rdf" "http://www.w3.org/1999/02/22-rdf-syntax-ns#"`, `"ps" "http://vocab.belgif.be/ns/publicservice#"` and `"vcard" "http://www.w3.org/2006/vcard/ns#"` are missing.

config/resources/lpdc.lisp
```diff
@@ -0,0 +1,58 @@
...
+                (:description :language-string-set ,(s-prefix "dct:description"))
+                (:additional-description :language-string-set ,(s-prefix "lpdcExt:additionalDescription"))
+                (:keyword :language-string-set ,(s-prefix "dcat:keyword"))
+                (:regulations :language-string-set ,(s-prefix "lpdcExt:regulation"))
+                (:exceptions :language-string-set ,(s-prefix "lpdcExt:exception"))
+                (:start-date :datetime ,(s-prefix "schema:startDate"))
+                (:end-date :datetime ,(s-prefix "schema:endDate"))
+                (:created :datetime ,(s-prefix "dct:created"))
+                (:modified :datetime ,(s-prefix "dct:modified"))
+                (:product-id :string ,(s-prefix "schema:productID")))
```
Riadabd, line 13: I was thinking of having consistent spacing in our migration and resource code in general. For example,
```
((:name                   :language-string-set  ,(s-prefix "dct:title"))
 (:description            :language-string-set  ,(s-prefix "dct:description"))
 (:additional-description :language-string-set  ,(s-prefix "lpdcExt:additionalDescription"))
 (:keyword                :language-string-set  ,(s-prefix "dcat:keyword"))
 (:regulations            :language-string-set  ,(s-prefix "lpdcExt:regulation"))
 (:exceptions             :language-string-set  ,(s-prefix "lpdcExt:exception"))
 (:start-date             :datetime             ,(s-prefix "schema:startDate"))
 (:end-date               :datetime             ,(s-prefix "schema:endDate"))
 (:created                :datetime             ,(s-prefix "dct:created"))
 (:modified               :datetime             ,(s-prefix "dct:modified"))
 (:product-id             :string               ,(s-prefix "schema:productID")))
```

Given that it's data defined in Lisp code, I find it visually better to have spacing in order to notice the separation better. Do you think it would look better this way?

config/resources/repository.lisp
Windvis: I don't think they are used here? I only added the ones that were used in the file, but maybe I missed some.

config/resources/lpdc.lisp
cecemel: agree with @Windvis

Riadabd (approved):

PR #409, opened 2023-05-08
[LPDC] Remove empty generator values

Windvis:
Prefilling the fields with a space, or an empty string has little value and bypasses the "required" validator potentially resulting in bad data.


Windvis:
Oh, wow, I totally forgot about this one. Thanks!

PR #411, opened 2023-05-08
Add report for LPDC instances stuck in publishing

Riadabd:
(Addresses DL-5087) Adds report for LPDC instances that are sent but stuck in publishing (includes public services and tombstones).


cecemel (changes requested):
I think we need a bit more info:
the name (label) of the bestuurseenheid + classificatie (label)
Else not helpful to troubleshoot in a later stage. 

For the tombstone, getting this info is a bit annoying, and brittle, 
you will have to parse the ?graph name extract the uuid, and go back eenheid like this.

cecemel:
@Riadabd and where possible also the NL title. easier to troubleshoot again

cecemel:
@Riadabd  I added some commits, please check em (and test next time :-) )

PR #412, opened 2023-05-10
[LPDC] Implement the concept label and filtering functionality

Windvis:


config/migrations/2023/lpdc/20230512123030-add-concept-display-configurations.sparql
```diff
@@ -0,0 +1,42 @@
...
+  # find the concepts which have no config yet
+  GRAPH <http://mu.semte.ch/graphs/public> {
+    ?concept a lpdcExt:ConceptualPublicService
+  }
+
+  FILTER NOT EXISTS {
+    GRAPH ?graph {
+      ?concept lpdcExt:hasConceptDisplayConfiguration ?configUri
+    }
+  }
```
Windvis, line 38: Something is still missing here I think. It also adds new ConceptDisplayConfigurations even if one exists for that concept / bestuurseenheid combo already.

config/migrations/2023/lpdc/20230512120000-add-concept-display-configurations-for-instanced-concepts.sparql
```diff
@@ -0,0 +1,33 @@
...
+      lpdcExt:conceptInstantiated true ;
+      dct:relation ?eenheid .
+  }
+}
+WHERE {
+  GRAPH ?graph {
+    ?instance a cpsv:PublicService;
+      dct:source ?concept;
+      pav:createdBy ?eenheid.
+  }
```
Windvis, line 23: If a bestuurseenheid has multiple instances, the same concept will be returned here, which means multiple Configs would be inserted in the DB. The current configId kind of bypasses that since the uri would be the same for each concept duplicate.

I tried using a subquery to only select the unique concepts and the end result seemed the same, not sure if that's the correct solution here.

config/migrations/2023/lpdc/20230512123030-add-concept-display-configurations.sparql
Windvis, line 1: It is quite a heavy query. I assume it will take even longer on PRD since there are most likely more concepts there? What do we do if it times out?

config/migrations/2023/lpdc/20230512123030-add-concept-display-configurations.sparql
Windvis: I went for a different solution which sidesteps this issue.

config/migrations/2023/lpdc/20230512120000-add-concept-display-configurations-for-instanced-concepts.sparql
Windvis: I solved this in a different way that sidestepped this issue.

config/migrations/2023/lpdc/20230512160000-add-concept-display-configurations.sparql
```diff
@@ -0,0 +1,36 @@
...
+    ?configUri a lpdcExt:ConceptDisplayConfiguration ;
+      mu:uuid ?configId ;
+      lpdcExt:conceptIsNew true ;
+      lpdcExt:conceptInstantiated false ;
+      dct:relation ?eenheid .
+  }
+}
+WHERE {
+  VALUES ?eenheidClassificatie {
+    <http://data.vlaanderen.be/id/concept/BestuurseenheidClassificatieCode/5ab0e9b8a3b2ca7c5e000001> # gemeente
```
cecemel, line 20: @Windvis same remark, double check the list thanks!

cecemel:
@Windvis i tested the migration it seems to run fine on prod db

cecemel:
@Windvis tool samples from the resulting migration, looked like it did what it supposed to be

config/migrations/2023/lpdc/20230512160000-add-concept-display-configurations.sparql
Windvis: We also need to add "Provincies" to the list here. Do I create a separate migration for each type to limit the chance of timeouts?

config/migrations/2023/lpdc/20230512160000-add-concept-display-configurations.sparql
```diff
@@ -0,0 +1,37 @@
...
+}
+WHERE {
+  VALUES ?eenheidClassificatie {
+    <http://data.vlaanderen.be/id/concept/BestuurseenheidClassificatieCode/5ab0e9b8a3b2ca7c5e000000> # provincie
+    <http://data.vlaanderen.be/id/concept/BestuurseenheidClassificatieCode/5ab0e9b8a3b2ca7c5e000001> # gemeente
+  }
+
+  ?eenheid a besluit:Bestuurseenheid;
+    mu:uuid ?eenheidId;
+    besluit:classificatie ?eenheidClassificatie.
```
Riadabd, line 26: Minor nitpick: The other triples in the queries have a space between the object and either `.` or `;` whereas these don't.

Riadabd (approved):
Looks good! Tested with the new resources + mu-auth combinations as well.

config/migrations/2023/lpdc/20230512160000-add-concept-display-configurations.sparql
```diff
@@ -0,0 +1,37 @@
...
+PREFIX lpdcExt: <https://productencatalogus.data.vlaanderen.be/ns/ipdc-lpdc#>
+PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
+PREFIX pav: <http://purl.org/pav/>
+
+INSERT {
+  GRAPH ?graph {
+    ?concept lpdcExt:hasConceptDisplayConfiguration ?configUri .
+    ?configUri a lpdcExt:ConceptDisplayConfiguration ;
+      mu:uuid ?configId ;
+      lpdcExt:conceptIsNew true ;
```
Windvis, line 13: Double checked with Stefanie. This needs to be set to false in order to reduce the noise this will otherwise cause. The risk is that users miss some new concepts, but it would have been the same if this feature wasn't implemented.

config/migrations/2023/lpdc/20230512160000-add-concept-display-configurations.sparql
cecemel: yeah how many provincies do we have these days again? :-)

PR #420, opened 2023-05-23
Stadsvernieuwing fase 2 - feeback round

aatauil:
This PR makes small changes to the stadsvernieuwing fase 2 formulier. Nothing too special.

To test:
- only backend changes, setup stack and checkout this branch
- login as deinze
- create new stadsvernieuwing projectsubsidie and fill in first step
- In the second step you need to go over all changes and check if everything saves correctly

The changes include:
- text changes/additions
- increase character count for some fields
- remove required validation on  relevante bijlage (bottom of form)
- change indienen tot date to 19 september

For detailed overview see DL- 5138

### DIFF
--- a/config/subsidy-application-management/forms/urban-renewal/application-phase2/versions/20221218090000/form.ttl
+++ b/config/subsidy-application-management/forms/urban-renewal/application-phase2/versions/20221218090000/form.ttl
@@ -316,7 +316,7 @@ ext:samenvattingPropertyGroup
       form:validations
                         [ a                 form:MaxLength ;
                           form:grouping     form:MatchEvery ;
-                          form:max          "700" ;
+                          form:max          "5000" ;
                           sh:resultMessage  "Max. karakters overschreden." ;
                           sh:path           ext:samenvattingSummaryDescription
                         ],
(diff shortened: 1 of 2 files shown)


config/subsidy-application-management/forms/urban-renewal/application-phase2/versions/20221218090000/form.ttl
```diff
@@ -869,7 +869,7 @@ ext:projectOnderdeelPg a form:PropertyGroup;
             form:validations
                               [ a                 form:MaxLength ;
                                 form:grouping     form:MatchEvery ;
-                                form:max          "500" ;
+                                form:max          "1000" ;
                                 sh:resultMessage  "Max. karakters overschreden." ;
                                 sh:path            lblodSubsidie:projectOnderdeelStadListingUnitAandeel
                               ];
```
Riadabd, line 875: The character limit works as expected, but the counter doesn't show up (same for `Publieke partner` and `Private Partner`). Is it because it's in a table?

Windvis: Yea, some things are hidden when a field is displayed in a table to conserve space. It's honestly kind of weird to that users would input that many characters in a table. Especially when it's a normal input field.

PR #424, opened 2023-06-07
migration moving the thematische subsidie from public to org graphs

aatauil:
Currently thematische subsidies is visible for everyone (public graph). This PR removes it from public graph and puts it in org graphs of those that should be able to see the subsidie.

To test:
- setup stack en make sure migration has run
- login as different bestuurseenheden and check that the subsidie is only available for the ones that are in the list.
( you can see permissions list in migration file itself )


Riadabd (approved):
Looks good!

PR #425, opened 2023-06-15
Enable the Virtuoso compatibility option in mu-auth

Windvis:
This solves an issue where some state isn't properly updated in the database due to a Virtuoso bug.

More information: https://github.com/mu-semtech/mu-authorization/#database-compatibility

I only tested it in combination with `redpencil/virtuoso:1.1.0` so it still needs to be verified that it works on the current image as well. We potentially need to merge #426 as well.

DL-5205


cecemel:
Closing

PR #438, opened 2023-07-18
bumb bg initiator to reduce number of containers

nbittich:


cecemel:
@nbittich thanks, we'll do it gradually

PR #455, opened 2023-09-15
Share subsidy attachments over delta-sync

wolfderechter:
## Description

Simple PR which extends the allowed accounts and delta-producer of subsidies so the subsidy attachments can get shared over delta-sync. 

 ## Type of change

 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [x] New feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
 - [ ] Other

 ## Related services

 - delta-files-share

 ## How to test

 ## Links to other PR's

TBD

PR #456, opened 2023-09-20
Extend subsidy delta producer

Riadabd:
This PR begins the process of extending the subsidy export producer so that we can consume this data in `app-subsidiedatabank`.


Riadabd:
In `/home/r2d2/Code/rpio/lblod/app-digitaal-loket/config/subsidy-application-management/forms/ukraine-slaapplekken/proposal/versions/20220509100012/form.json`:
```
{
  "s-prefix": "lblodSubsidie:accountabilityTable",
  "properties": [
    {
      "s-prefix": "lblodSubsidie:accountabilityEntry",
      "properties": [
        "ukraine:address",
        "ukraine:bedroomCount",
        "ukraine:sharedInvoice",
        "ukraine:hasFiles",
        "dct:created"
      ]
    }
  ]
}
```

The block is not available on the production database, but it may become available there at any point. I'm keeping this comment for the future to add to `export.json` once we can query the data and find the relevant types.

Riadabd:
`FileDataObject` was removed from this PR because it's handled in [another PR](https://github.com/lblod/app-digitaal-loket/pull/455).

cecemel:
@Riadabd you consider this ready, just to make sure. 
Also, any chances you tested this with a healing job on prod data? We might need to use the updated publication-graph-maintainer, because better performance. This will be a PR on this PR

aatauil:
@Riadabd Encountered a somewhat annoying issue. The properties of contactPoint (firstname lastname..) are not directly linked but are instead linked through an intermiddiate subject uri 
```
<contactpoint/123> predicate <form/data/123>
<form/data/123> foaf:name: "name"
<form/data/123> foaf:tel: "0123435"
```

Which means that searching for firstName on contactPoint directly will not discover the data we want to be exported. This is just an example of where I found that it happens. <form/data/123> also doesnt have a rdf:type or mu:uuid which complicates things further. 

Riad could you investigate further?

cecemel:
> @Riadabd Encountered a somewhat annoying issue. The properties of contactPoint (firstname lastname..) are not directly linked but are instead linked through an intermiddiate subject uri
> 
> ```
> <contactpoint/123> predicate <form/data/123>
> <form/data/123> foaf:name: "name"
> <form/data/123> foaf:tel: "0123435"
> ```
> 
> Which means that searching for firstName on contactPoint directly will not discover the data we want to be exported. This is just an example of where I found that it happens. <form/data/123> also doesnt have a rdf:type or mu:uuid which complicates things further.
> 
> Riad could you investigate further?

@Riadabd @aatauil thanks; i might just say trivial things but in the example from achraf it would be a matter of adding a type 'Contact' to  <form/data/123>

aatauil:
@cecemel Yeah but a migration like that would only add the type to **existing** <form/data/123> subjects. When a **new form** gets submitted the data will be stored in a new subject: <form/data/456789>.  So we might need to create a migration for existing ones ofc but also update subsidy-application-management-service to add a type whenever it creates <form/data/123> or <form/data/456789>. 

Not sure this happens in subsidy-application-management-service and even why an intermidiate subject gets added so thats why we need to analyse a bit

cecemel:
> @cecemel Yeah but a migration like that would only add the type to **existing** <form/data/123> subjects. When a **new form** gets submitted the data will be stored in a new subject: <form/data/456789>. So we need to create a migration for existing ones ofc but probably also update subsidy-application-management-service to add a type whenever it creates <form/data/123> or <form/data/456789>.
> 
> Not sure this happens in subsidy-application-management-service and even why an intermidiate subject gets added so thats why we need to analyse a bit

@aatauil @Riadabd  yep that's one of these undetected legacy buggy bits. But
1. I wouldn't block the PR for that.
2. There are multiple ways to tackle these, i would only focus also on the forms that are still active.  

# solution 1: update generators (most generic)
 -> Add the `rdf:type` here: https://github.com/lblod/app-digitaal-loket/blob/master/config/subsidy-application-management/forms/urban-renewal/application/versions/20221218090000/form.json#L14
   -> use an init generator that once the form loads, the resource is created. 
      -> see e.g. https://github.com/lblod/ember-submission-form-fields/blob/d46b9d199ab22899456ebb0e410f5902a29da09f/tests/dummy/public/test-forms/generators/form.ttl#L18 on how to add these.
             -> I had a brief look at the implementation; It might be buggy; we use generators on hasMany; i don't think we use init generators anywhere. From the implementation; it looks like every time the form is rendered, the generator will create a resource. That's not what we want; we need to first check whether there is any resource matches the output from the generator (and maybe cardinality) and if not; run the generator. As a starting point look here https://github.com/lblod/ember-submission-form-fields/blob/d46b9d199ab22899456ebb0e410f5902a29da09f/addon/components/rdf-form.js#L32

# solution 2: simplest but verbose and custom
 - Another approach would be to use/abuse the extractors and add those types there. 
 - you will still need to add the `rdf:type` here: https://github.com/lblod/app-digitaal-loket/blob/master/config/subsidy-application-management/forms/urban-renewal/application/versions/20221218090000/form.json#L14

# solution 3: most complex
update subsidy  management so it can deal with an updated spec of the form.json; i.e. by adding a field `'type': 'schema:ContactPoint'`
This is my least favorite option, since complex and not very re-usable, but it's a good opportunity to have a deeper dive in this complex service.

cecemel:
@Riadabd @aatauil what's the state here?

aatauil:
@cecemel short summary. Tried solution 1 & 2 but encountered issues or was hacky. Feel free to ping for specifics. Solution 3 seems best but will take time. I am suggesting the following.

### Solutions 4: add ext:formDataNode rdf:type
Add a form-data type for future forms. `<form-data/node/123> a ext:formDataNode`. The `<form-data/node/123>` gets generated in the submission-form-helpers library. Specifically here:
https://github.com/lblod/submission-form-helpers/blob/bf22a5527a642d0b89724926f185ece6c3c9d35d/src/triples-for.js#L35-L54
We can push the extra rdf:type triple in the list for future forms
```js
triples.push(
      new Statement(
        new NamedNode(URI_TEMPLATE + uuidv4()),
        NODE("rdf:type"),
        NODE("ext:formDataNode"),
        sourceGraph
      )
    );
```

Add migration updating past forms:
```sparql
INSERT {
  ?formDataNode a ext:formDataNode
}
WHERE {
  ?formDataNode ?p ?o .
  FILTER(STRSTARTS(STR(?formDataNode), "http://data.lblod.info/form-data/nodes/"))
}
```

then we can export by using that type:
```jsonc
"properties": [
  "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
  "http://mu.semte.ch/vocabularies/core/uuid",
  "http://schema.org/email",
  "http://schema.org/telephone",
  "http://xmlns.com/foaf/0.1/firstName",
  "http://xmlns.com/foaf/0.1/familyName"
],
"graphsFilter": [
  "http://mu.semte.ch/graphs/organizations/.*/LoketLB-subsidies"
],
"type": "http://mu.semte.ch/vocabularies/ext/formDataNode" // we now have a type we can use
```

Riadabd:
@aatauil Solution 4 seems like a good approach to take, but it leaves out types like `ContactPoint` and `PoliticalReferenceContactPoint`; `ContactPoint` already has a predefined (`http://schema.org/ContactPoint`) type, and I gave `PoliticalReferenceContactPoint` a type of `http://mu.semte.ch/vocabularies/ext/PoliticalReferenceContactPoint`.

We can consider contact points to be form data in the context of a subsidy application, but doing so may cause issues down the line in case a decision is made to have a consistent way of re-using contact information.

Riadabd:
I noticed that bank account fields created as part of the `Betaling` section in the application forms do not have an associated type. Normally this should be `http://schema.org/BankAccount`, but a look up in the database yields a form data URI with no type (i.e., `<http://data.lblod.info/id/application-forms/6554B7A92DC683695EB55412> <http://schema.org/bankAccount> <http://data.lblod.info/form-data/nodes/1a370b7c-b64b-4ef1-ad6e-708cffeaafca>`).

Expanding `http://data.lblod.info/form-data/nodes/1a370b7c-b64b-4ef1-ad6e-708cffeaafca` gives `http://purl.org/dc/terms/hasPart` (for the bevestingsbrief) and `http://schema.org/identifier` (for the IBAN number) properties.

PR #472, opened 2023-10-30
OP has implemented RO code. using this one in Loket instead

nbittich:
DL-5462

### DIFF
--- a/config/reasoner/op2dl/main/mapping.n3
+++ b/config/reasoner/op2dl/main/mapping.n3
@@ -33,13 +33,14 @@ PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
   ?s a besluit:Bestuurseenheid
 }.
 
-{
-  ?s a ere:RepresentatiefOrgaan
-}=> {
-  ?s
-    a besluit:Bestuurseenheid;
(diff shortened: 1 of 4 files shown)


config/reasoner/op2dl/main/mapping.n3
```diff
@@ -33,13 +33,14 @@ PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
...
-  ?s
-    a besluit:Bestuurseenheid;
-    besluit:classificatie <http://data.vlaanderen.be/id/concept/BestuurseenheidClassificatieCode/36372fad-0358-499c-a4e3-f412d2eae213>.
-}.
+# this rule should no longer be needed as OP has a classification for RO now
+# {
+#  ?s a ere:RepresentatiefOrgaan
+# }=> {
+#  ?s
+#    a besluit:Bestuurseenheid;
```
claire-lovisa, line 41: I think that we still need to add the `besluit:Bestuurseenheid` type because ACMIDM needs it for the log in ? Not 100% sure, but worth checking

claire-lovisa (commented):
I think that many other things need to be updated, for example some `BesluitDocumentType` seem to be decidable (`<http://lblod.data.gift/vocabularies/besluit/decidableBy>`) by the old RO classification so those links would need to be updated in the DB.
I'm also not sure about the consequence of this change on submitted data, if it's then published and consumed elsewhere it could cause some issues but I don't have enough knowledge on Loket to be certain about it, @cecemel will probably have more insights

PR #480, opened 2023-11-14
Extend fietssubsidie configuration

Riadabd:
This PR uses a [previous one](https://github.com/lblod/app-digitaal-loket/pull/456) as a reference with the goal of extending the configuration one subsidy at a time; this is to reduce the possible number of errors that may arise from exporting subsidies all at once.

### DIFF
--- a/config/delta-producer/subsidies/export.json
+++ b/config/delta-producer/subsidies/export.json
@@ -117,7 +117,15 @@
         "http://lblod.data.gift/vocabularies/subsidie/plan-samenleven/usesSocialImpactBound",
         "http://lblod.data.gift/vocabularies/subsidie/bicycle-infrastructure#estimatedCostTable",
         "http://lblod.data.gift/vocabularies/subsidie/bicycle-infrastructure#objectiveTable",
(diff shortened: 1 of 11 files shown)


Riadabd:
After checking the three steps of the fietssubsidie, the only data not showing up are the fields that do not have a `type` predicate. More discussion around this topic can be found [here](https://github.com/lblod/app-digitaal-loket/pull/456).

Riadabd:
I tested the PR with fietssubsidies from different gemeenten, and the subsidy-specific data (documents, names, numbers, etc.) were correctly consumed by `app-subsidiedatabank`. `ContactPoint` was not added to the export as it is already part of [another PR](https://github.com/lblod/app-digitaal-loket/pull/485).

wolfderechter (approved):
Tried running it with the ContactPoint exports as well and worked perfectly for me!

aatauil:
Converting to draft to prevent accidental merge. This PR looks good but should only be merged after the newly created subsidieforms get the correct types attached. The config extend through extractor is still a WIP

PR #485, opened 2023-11-23
Extend LEKP 1.0 configuration

Riadabd:
This PR uses a [previous one](https://github.com/lblod/app-digitaal-loket/pull/456) as a reference; the goal here is to extend the LEKP 1.0 subsidy producer configuration so that all data shows up on subsidiedatabank. The PR also adds custom types to specific entities that currently have none inside the database.


wolfderechter:
Ran this last week and worked as expected :ok_hand:

aatauil:
Closed in favour of the following: https://github.com/lblod/app-digitaal-loket/pull/523. We went with a different method (using extractors)

PR #489, opened 2023-11-27
Optimising VDDS query; we have a winner

cecemel:
This is another optimization of the where clause in VDDS for berichten. It's decent speed. Why? Because.... no idea


cecemel:
@benjay10 fyi

PR #493, opened 2023-12-05
Fix: Messages via Controle from ABB not correctly saved and distributed

benjay10:
Saving a message in Loket from ABB (via Controle) was not done correctly. Sender and recipient were not always correct. In combination, the VDDS was not triggering correctly on those messages either. This meant that vendors did not see those messages in their vendor graph.

Fixed by adding config to the VDDS to also trigger on those messages from ABB and created by the Loket frontend. A new version of the frontend also fixes the sender and recipients for messages.

**TODO:**

* Review and release:

  - https://github.com/lblod/frontend-loket/pull/349
  - https://github.com/lblod/frontend-loket/pull/350

* Bump Loket frontend in this PR

### DIFF
--- a/config/vendor-data/subjectsAndPaths.js
+++ b/config/vendor-data/subjectsAndPaths.js
@@ -149,6 +149,8 @@ export const subjects = [
            } UNION {
              ?conversatie ?pc ?co .
            } UNION {
+             ?bericht ?pa ?ca .
+           } UNION {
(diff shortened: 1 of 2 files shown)


cecemel:
@benjay10 seems doing fine. I also fixed a bug introducing the union contstruct..

cecemel (approved):

PR #494, opened 2023-12-05
Set report cron pattern to a past date

Riadabd:
## Issue Number

DL-5497

## Issue Description

The following reports:
* List of proposals of fiets subsidies for checking deadline gemeenten
* List of proposals of fiets subsidies
* List of pacts for LEKP 2.1
* Subsidie Stadsvernieuwing Aanvraag (Fase 2) Report
are no longer needed and should be disabled.

## How to Test

The PR sets the cron pattern to `1/1/2023` to disable the reports from running. In order to test this, set the cron pattern time section, while keeping the date one, to your local time and observe the report generation service (`docker compose logs -ft report-generation`) to see if the reports are running.

> **_NOTE:_** The cron pattern time needs to be set to one hour before your local time since Belgium is UTC+1 at the moment.


cecemel:
@Riadabd i think it might be more clear to add these to archive folder
https://github.com/lblod/app-digitaal-loket/tree/master/config/reports/archive
and remove their import from https://github.com/lblod/app-digitaal-loket/blob/master/config/reports/index.js

PR #519, opened 2024-01-29
[DL-5624] Add a warning validation to the "Datum zitting" fields

Windvis:
This adds the `DateInPast` constraint to all "Datum zitting" fields and configures them as warnings.

This is the same change as https://github.com/lblod/manage-submission-form-tooling/pull/43 but directly to the current active form.

~~Needs https://github.com/lblod/frontend-loket/pull/358 to work properly.~~ (Part of the bundled v0.90.0 release now)

**To test:**
- start the stack
- access the bundled frontend and go to the toezicht module 
> The bundled frontend only works if you access it on localhost (or https), because of the [crypto.randomUUID()](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) util that is used by EmberData)
- create a new submission
- select a type of submission that has a `Datum zitting/ *` field. There are multiple versions of this field with a different label
- input a date in the future
- verify that the warning validation is displayed. It should look like this:
image

- verify that the form can be submitted, even if the selected date is in the future

### DIFF
--- a/config/semantic-forms/20240110093200-forms.ttl
+++ b/config/semantic-forms/20240110093200-forms.ttl
@@ -314,7 +314,12 @@ fields:3dd6ed93-40f7-4d70-a6cb-f4de53dc8bfb a form:Field ;
       [ a form:ValidDateTime ;
         form:grouping form:MatchEvery ;
         sh:path ( [ sh:inversePath prov:generated ] dct:subject [ sh:inversePath besluit:behandelt ]  prov:startedAtTime ) ;
-        sh:resultMessage "Geef een geldige datum en tijd op."] ;
+        sh:resultMessage "Geef een geldige datum en tijd op."],
+      [ a form:DateInPast ;
+        form:grouping form:MatchEvery ;
+        sh:path ( [ sh:inversePath prov:generated ] dct:subject [ sh:inversePath besluit:behandelt ]  prov:startedAtTime ) ;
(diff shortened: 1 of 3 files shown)


config/semantic-forms/20240110093200-forms.ttl
```diff
@@ -314,7 +314,12 @@ fields:3dd6ed93-40f7-4d70-a6cb-f4de53dc8bfb a form:Field ;
       [ a form:ValidDateTime ;
         form:grouping form:MatchEvery ;
         sh:path ( [ sh:inversePath prov:generated ] dct:subject [ sh:inversePath besluit:behandelt ]  prov:startedAtTime ) ;
-        sh:resultMessage "Geef een geldige datum en tijd op."] ;
+        sh:resultMessage "Geef een geldige datum en tijd op."],
+      [ a form:DateInPast ;
+        form:grouping form:MatchEvery ;
+        sh:path ( [ sh:inversePath prov:generated ] dct:subject [ sh:inversePath besluit:behandelt ]  prov:startedAtTime ) ;
+        sh:resultMessage "De gekozen datum ligt in de toekomst. Gelieve deze na te kijken." ;
```
Windvis, line 321: The message is still pending, feedback requested in Jira

Windvis: We're going for "Let op: datum ligt in de toekomst. Controleer uw keuze." instead, so that should be updated here.

PR #520, opened 2024-01-30
LEKP 1.0 - 2021 - Delete unneeded concept-schemes & form properties

aatauil:
## ID
 DGS-144

 ## Description

While updating extractor I stumbled across some issues because too many properties where defined in the form its json file. 

- too many properties cause unstable extractor behaviour (specifically adms:status)
- concept-schemes defined in meta get fetched but are never used

 ## Type of change

 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [ ] New feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
 - [x] Other

 ## How to test

Best to test both with local and with remote data. 
- setup the app
- create a LEKP 1.0 - 2021 subsidy
- fill in the form and all its values
- save, reload and navigate around a bit 
- check if all data is loading in correctly and nothing odd is happening

### DIFF
--- a/config/subsidy-application-management/forms/climate/step-submit-opvolgmoment-2024/versions/20231116154454/form.json
+++ b/config/subsidy-application-management/forms/climate/step-submit-opvolgmoment-2024/versions/20231116154454/form.json
@@ -48,19 +48,5 @@
         ]
       }
     ]
-  },
-  "meta": {
-    "schemes": [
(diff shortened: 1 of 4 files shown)


wolfderechter (approved):
Everything seems to be working for me

PR #526, opened 2024-02-09
Bump delta-producer-publication-graph-maintainer

cecemel:
This should fix the broken mandatarissen and leidinggevenden live sync. (DL-5584)


cecemel:
@Riadabd fyi

PR #532, opened 2024-02-28
Remove unused frontend env vars

Riadabd:
Remove unused frontend environment variables from `docker-compose.dev.yml`.


Riadabd:
The external module card behavior is unchanged after removing these variables.

PR #543, opened 2024-03-14
[DL-5618] Add a new "Virus scan" report

Windvis:
This report will generate a list of all the files which are considered "malicious" or where the status of the scan is unknown.

**Test instructions**
- check out the branch and start the stack
- add a port mapping to the report generation service so you can make direct API calls to it:
```yml
  report-generation:
    ports:
      - "8080:80"
```
- restart the `report-generation` service if the stack was already running (`drc restart report-generation` / `drc up -d report-generation` in case of the port mapping)
- Upload a [malicious test file](https://www.eicar.org/download-anti-malware-testfile/) in one of the file upload fields in Loket. The easiest is to go to the `Toezicht` module, create a new submission and upload it there.
- (optional) check the virus scan service logs to see if the file was scanned
- trigger the report generation with an [API call](https://github.com/lblod/loket-report-generation-service?tab=readme-ov-file#manually-trigger-reports) using `virusScanReport` as the name
- verify that a file is generated starting with the following name `virus-scan-report-[uuid].csv` and that it includes an entry for the malicious test file

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@
    - Adds `vendor-management` as a new delta-stream
  - Add server-specific configurations for `delta-producer-background-jobs-initiator` and `delta-producer-publication-graph-maintainer` (DL-5752)
    - Resolves merge conflict issues when adding a new delta stream or editing an existing one
- - Add `restart: always` for the `virus-scanner` service
+ - Add a new "Virus scan" report (DL-5618)
 ### Deploy Notes
  - Remove the frontend `v0.90.3` image override from `docker-compose.override.yml`
  - Update the controle image to `v0.91.1-controle` in the `docker-compose.override.yml` file
@@ -38,7 +38,7 @@
  - Add `"key": "<producer_key>"` for the new `vendor-management` delta stream.
 #### Docker Commands
  - `drc up -d loket controle migrations mocklogin vendor-data-distribution enrich-submission prepare-submissions-for-export virus-scanner berichtencentrum-sync-with-kalliope`
- - `drc restart subsidy-applications-management dispatcher deltanotifier delta-producer-background-jobs-initiator delta-producer-publication-graph-maintainer delta-producer-dump-file-publisher jobs-controller delta-producer-publication-graph-maintainer-subsidies migrations resource cache`
+ - `drc restart subsidy-applications-management dispatcher deltanotifier delta-producer-background-jobs-initiator delta-producer-publication-graph-maintainer delta-producer-dump-file-publisher jobs-controller delta-producer-publication-graph-maintainer-subsidies migrations resource cache report-generation`
 ## 1.94.0 (2024-02-19)
 ### Subsidies
  - Add new stadsvernieuwing - conceptsubsidie || Oproep 2024 reeks (DGS-154)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -22,7 +22,7 @@
    - Adds `vendor-management` as a new delta-stream
  - Add server-specific configurations for `delta-producer-background-jobs-initiator` and `delta-producer-publication-graph-maintainer` (DL-5752)
    - Resolves merge conflict issues when adding a new delta stream or editing an existing one
- - Add `restart: always` for the `virus-scanner` service
```
Windvis, line 25: This doesn't really need to be in the changelog, since it's just an internal technical thing.

config/reports/virusScanReport.js
```diff
@@ -0,0 +1,47 @@
...
+    ?analysis a stix:MalwareAnalysis ;
+      stix:analysis_started ?scanDate ;
+      stix:sample_ref ?file ;
+      stix:result ?scanResult .
+      FILTER (?scanResult IN ("malicious", "unknown"))
+
+      ?file nfo:fileName ?fileName .
+  }
+}
+ORDER BY DESC(?scanDate)`;
```
Riadabd, line 31: Small indentation nitpick; I'm not sure if Github will display this properly, but in case it doesn't, I meant something akid to [this](https://github.com/lblod/app-digitaal-loket/blob/master/config/reports/submissionsReport.js#L21-L41).

```suggestion
    const queryString = `
      PREFIX stix: <http://docs.oasis-open.org/cti/ns/stix#>
      PREFIX nfo: <http://www.semanticdesktop.org/ontologies/2007/03/22/nfo#>
      PREFIX nie: <http://www.semanticdesktop.org/ontologies/2007/01/19/nie#>
      
      SELECT DISTINCT ?file ?fileName ?scanResult ?scanDate WHERE {
        GRAPH ?g {
          ?analysis a stix:MalwareAnalysis ;
            stix:analysis_started ?scanDate ;
            stix:sample_ref ?file ;
            stix:result ?scanResult .
            FILTER (?scanResult IN ("malicious", "unknown"))
      
            ?file nfo:fileName ?fileName .
        }
      }
      ORDER BY DESC(?scanDate)
    `;
```

CHANGELOG.md
Riadabd: Yes, you're right indeed.

Riadabd (approved):
Besides the small style nitpick in the report, all works great!

* `virus-scanner` logs show the service scanning uploaded files.
* The virus scan report is successfully generated with the API call and contains only the malicious text file (from amongst a couple of regular mock files).

config/reports/virusScanReport.js
cecemel, line 31: Would it be more 'in your face' if you first orde by status (malicious on top) and then by date?

Windvis: So something like `ORDER BY ?scanResult DESC(?scanDate)`?

Windvis: updated the indentation

cecemel: yeah

Riadabd (approved):
Good to go!

PR #545, opened 2024-03-15
Person impersonation

cecemel:
This PR adds the impersonation service.
It currently reuses the `LoketAdmin` role to allow users to impersonate other accounts.

We've taken the opportunity to update the MockAccount of ABB, to a real Admin mock account.
This will be the only mock account you can use for the impersonation service.

Note: for testing, we'll need to ask the QA team to double-check the dashboard and vendor login to ensure it still works properly.
Note 2: see https://github.com/lblod/impersonation-service/pull/2

Context: preparative work for DL-5633


Windvis (commented):
Tested this locally by logging in as the "Mock Agentschap Binnenland bestuur (admin)" user and sending some fetch requests from the console to create / retrieve / delete the impersonation.

I was able to retrieve data for the "Gemeente Aalst" account 👍 while being logged in as the admin user.

One thing that could be improved is that `GET /impersonations/current` doesn't return a 4xx response if there is no impersonation. Instead it returns the following response:

```
{"links":{"self":"/impersonations/current"},"data":{"type":"impersonations"}}
```

So a json:api like response, but without an id. Ideally a 400 response would be returned, similar to how it works in the [acmidm-login service](https://github.com/lblod/acmidm-login-service/blob/266314133ea9761a35e471706e49c3dfef254181/app.js#L148) and [mock-login service](https://github.com/lblod/mock-login-service/blob/0c80b5bdb48fb96bb146cd5e1483f6b8a985e0a4/web.rb#L168).

I also tested impersonating an account as a non-admin user, and it didn't create the impersonation, so the mu-auth config seems to work as expected 👍 With the caveat that it also didn't return a 4xx response, so it seemed like it did work before calling GET `/impersonations/current` to verify. That is less of an issue, since only admin users should be able to access the route that sets up the impersonations, so setting up the impersonation shouldn't ever fail in practice because of missing permissions.

<details>
<summary>Test API calls</summary>

Run these from the browser console, so the correct cookies are sent with the requests.

### Get the current impersonation
```js
fetch('/impersonations/current')
```

### Create an impersonation (for Gemeente Aalst)

```js
fetch('/impersonations', {
  method: 'POST',
  headers: {
    'Content-Type': 'application/vnd.api+json'
  },
  body: JSON.stringify({
    data: {
      type: "impersonations",
      relationships: {
        impersonates: {
          data: { type: 'resource', id: "9b875bc387960fd6efa0065bdff32877" }
        }
      }
    }
  }),
})
```

You can replace the id with a different account id if you want to impersonate a different user.

### Delete impersonation
```js
fetch('/impersonations/current', { method: 'DELETE' })
```

### Fetch toezicht data
```js
fetch('/submissions?filter[status][id]=79a52da4-f491-4e2f-9374-89a13cde8ecd,9bd8d86d-bb10-4456-a84e-91e9507c374c&include=form-data&page[number]=0&page[size]=20&sort=status.label,-sent-date,-modified', {
  headers: {
    'Content-Type': 'application/vnd.api+json'
  },
})
```

This call shouldn't return any data when not impersonating another user, but should once the impersonation is active.
</details>

docker-compose.yml
```diff
@@ -126,6 +126,8 @@ services:
       - "logging=true"
     restart: always
     logging: *default-logging
+  impersonation:
+    image: lblod/impersonation-service:0.2.0-rc.4
```
Riadabd, line 130: Should we add stuff like:
```yml
restart: always
logging: *default-logging
```
?

Riadabd: In addition to:
```
restart: no
```
inside `docker-compose.dev.yml`.

config/authorization/config.ex
```diff
@@ -21,16 +21,45 @@ defmodule Acl.UserGroups.Config do
...
+           ext:sessionRole ?session_role;
+           foaf:accountServiceHomepage <https://github.com/lblod/mock-login-service>.
+
+         ?person a foaf:Person;
+           foaf:account ?maybe_impersonated;
+           foaf:member/mu:uuid ?maybe_impersonated_group.
+       }
+       BIND(COALESCE(?maybe_impersonated_group, ?own_group) AS ?session_group)
+      }
+      LIMIT 1
```
Riadabd, line 58: Small style nitpick:
```suggestion
      PREFIX ext:       <http://mu.semte.ch/vocabularies/ext/>
      PREFIX mu:        <http://mu.semte.ch/vocabularies/core/>
      PREFIX foaf:      <http://xmlns.com/foaf/0.1/>
      PREFIX muAccount: <http://mu.semte.ch/vocabularies/account/impersonation/>

      SELECT DISTINCT ?session_group ?session_role WHERE {
        VALUES ?session_role {
          \"#{group_string}\"
        }

        VALUES ?session_id {
          <SESSION_ID>
        }

        {
          ?session_id ext:sessionGroup/mu:uuid ?own_group;
            ext:sessionRole ?session_role.
        }
        UNION
        {
          ?session_id muAccount:impersonates ?maybe_impersonated.
          ?maybe_impersonated a foaf:OnlineAccount;
            ext:sessionRole ?session_role;
            foaf:accountServiceHomepage <https://github.com/lblod/mock-login-service>.

          ?person a foaf:Person;
            foaf:account ?maybe_impersonated;
            foaf:member/mu:uuid ?maybe_impersonated_group.
        }

        BIND(COALESCE(?maybe_impersonated_group, ?own_group) AS ?session_group)
      } LIMIT 1
```

config/migrations/2024/20240314161912-ensure-mock-account-abb-is-admin.sparql
```diff
@@ -0,0 +1,30 @@
...
+  GRAPH ?g {
+    ?mock <http://mu.semte.ch/vocabularies/ext/sessionRole> ?role.
+  }
+  GRAPH ?j {
+    ?person <http://xmlns.com/foaf/0.1/familyName> ?name.
+  }
+}
+INSERT {
+  GRAPH ?g {
+    ?mock <http://mu.semte.ch/vocabularies/ext/sessionRole> "LoketAdmin", "LoketLB-vendorManagementGebruiker".
```
Riadabd, line 11: Is there a need to add "LoketLB-vendorManagementGebruiker" as a session role for ABB?

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Changelog
 ## Unreleased
+#### Frontend
+ - `v0.92.0` (DL-5790): https://github.com/lblod/frontend-loket/blob/development/CHANGELOG.md#v0920-2024-04-04
```
Windvis, line 4: The controle image no longer exists since we converted everything to runtime checks. Should we already remove the controle setup when deploying, or stick with it until we are sure the roles are set up correctly for the admin users?

docker-compose.dev.yml
```diff
@@ -7,7 +7,7 @@ services:
     ports:
       - "443:443"
   mocklogin:
-    image: lblod/mock-login-service:0.4.0
+    image: lblod/mock-login-service:0.5.0-beta.2
```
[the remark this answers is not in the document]
Riadabd: Yes, QA and PROD as well.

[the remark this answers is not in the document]
Riadabd: Maybe on PROD as well? I can see `mocklogin` there connected to the `controle` docker network.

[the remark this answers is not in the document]
cecemel: Once we have this feature, no more need for the the full controle setup in PROD

PR #555, opened 2024-04-03
Reports with too many double quotes

benjay10:
**[DL-5811]**

Some reports used a function called `getSafeValue` which was always redefined in those reports. This function provides safe property access that can more easily be written with the `?.` operation.

In most cases, it also calls a function `wrapInQuote` that wraps the values in the report with an extra double quote, which is no longer needed in newer versions of the `loket-report-generation-service`. In fact it is this function that is causing reports to be printed with too many quotes, making the reports unusable.

This PR removes those functions and relies on the better CSV handling in the service itself.

### DIFF
--- a/config/reports/ukraineSubsidyOproep1Report.js
+++ b/config/reports/ukraineSubsidyOproep1Report.js
@@ -1,5 +1,4 @@
 import { generateReportFromData, batchedQuery } from "../helpers.js";
-import { getSafeValue } from "./util/report-helpers";
 
 export default {
   cronPattern: "0 30 23 * * *",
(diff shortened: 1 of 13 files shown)


config/reports/ukraineSubsidyOproep1Report.js
```diff
@@ -87,26 +86,23 @@ export default {
     const queryResponse = await batchedQuery(queryString);
     const data = queryResponse.results.bindings.map((subsidie) => {
       return {
-        subsidieURI: getSafeValue(subsidie, "subsidie"),
-        modified: getSafeValue(subsidie, "modified"),
-        status: getSafeValue(subsidie, "status"),
-        bestuurseenheid: getSafeValue(subsidie, "bestuurseenheid"),
-        kbo: getSafeValue(subsidie, "kbo"),
```
Riadabd, line 94: Removing the double quotes from the KBO field causes the 0 on the left to be truncated (e.g., a kbo of `0xyz` becomes `xyz`) since I think it gets interpreted as a number.

Maybe we should keep the double quote wrapping on this one?

config/reports/ukraineSubsidyOproep1Report.js
```diff
@@ -87,26 +86,23 @@ export default {
...
-        classificatie: getSafeValue(subsidie, "classificatie"),
-        contactFirstName: getSafeValue(subsidie, "contactFirstName"),
-        contactLastName: getSafeValue(subsidie, "contactLastName"),
-        contactTelephone: getSafeValue(subsidie, "contactTelephone"),
```
Riadabd, line 98: Similar to the KBO field, removing the double quotes on the telephone number column leads to zeros being truncated. In addition, the column no longer stays "balanced"; in other words, without the double quotes, some phone numbers are aligned to the left (e.g., those that have spaces or contain the `/` character), and other phone numbers are aligned to the right (e.g., numbers like 04xxxxxx), which makes the column less readable than before.

benjay10:
@Riadabd I don't agree with adding back the quotes on those fields you mention. Double quotes are not part of a telephone number. Adding back double quotes around the telephone number is not even going to do anything. Excel will still evaluate that field as a number and truncate the first '0'. This was not the case when (by error) multiple extra quotes got introduced. In this case Excel saw the field as a string because of all the quotes and also rendered all the quotes as if they were part of the data. This is exactly what the end users didn't want.

I care about the correctness of the CSV document. This was previously not the case, but has gotten better in the last couple of versions of the report-generation-service. The way Excel formats the CSV document is important to the end user, but not at the cost of correctness of the CSV document in my opinion. The end user should now how to format columns in Excel to get the full telephone number to show. In LibreOffice, when opening the CSV document, you can choose the formatting of every specific column and when you select "Text" for the telephone numbers, it renders correctly. You can (I think) achieve something similar in Excel when you open an empty document and choose to import data from a text file. You should get a dialog like the one in LibreOffice with similar formatting setting.

Riadabd:
@benjay10 I tend to agree with you after reading your comment. My initial thoughts when writing my comments were based on how the end user's spreadsheet software would have after opening the csv file; however, you are indeed right that they should use capabilities provided by Excel/Libreoffice/other in order to format/sort however they see fit, rather than have the csv file formatted to a specific preference (instead of being correct).

I'll mark my comments as closed and approve the ticket (I had tested the PR on some data, and the double quotes were gone as expected).

PR #557, opened 2024-04-08
Use the ACM/IDM admin role

Windvis:
This updates the setup so it uses the `LoketLB-admin` role, (which is the one that ACM/IDM returns according to the login service logs on QA), for the impersonation feature.

There is already existing code that also uses the `LoketAdmin` role, but those are still required until ACM/IDM is enabled everywhere.

### DIFF
--- a/config/migrations/2024/20240408142300-mock-admin-correct-role.sparql
+++ b/config/migrations/2024/20240408142300-mock-admin-correct-role.sparql
@@ -1,0 +1,14 @@
+INSERT {
+  GRAPH ?g {
+    ?mock <http://mu.semte.ch/vocabularies/ext/sessionRole> "LoketLB-admin" .
+  }
+}
+WHERE {
+  VALUES ?mock {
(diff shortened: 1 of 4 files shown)


config/migrations/2024/20240408142300-mock-admin-correct-role.sparql
```diff
@@ -0,0 +1,14 @@
...
+}
+WHERE {
+  VALUES ?mock {
+    <http://data.lblod.info/id/account/3a91ff60-07c1-4136-ac5e-55cf401e0956>
+  }
+
+  GRAPH ?g {
+    ?mock <http://mu.semte.ch/vocabularies/ext/sessionRole> ?role .
+  }
+}
```
Riadabd, line 14: Small nitpick: Can you add a newline at the end?

Riadabd (approved):
Besides a small nitpick, all is good to go :+1:

PR #561, opened 2024-04-17
Impersonation refactoring

Windvis:
Needs this PR to work: https://github.com/lblod/impersonation-service/pull/6 (included as a feature image)

### DIFF
--- a/config/authorization/config.ex
+++ b/config/authorization/config.ex
@@ -22,43 +22,14 @@ defmodule Acl.UserGroups.Config do
   end
 
   defp sparql_query_for_access_role(group_string) do
-    """
-      PREFIX ext: <http://mu.semte.ch/vocabularies/ext/>
-      PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
-      PREFIX foaf: <http://xmlns.com/foaf/0.1/>
-      PREFIX muAccount: <http://mu.semte.ch/vocabularies/account/impersonation/>
-
-      SELECT DISTINCT ?session_group ?session_role WHERE {
-
-        VALUES ?session_role {
-           \"#{group_string}\"
-        }
-
-        VALUES ?session_id {
-           <SESSION_ID>
-        }
-
-       {
-          ?session_id ext:sessionGroup/mu:uuid ?own_group;
-                        ext:sessionRole ?session_role.
-       } UNION {
-
-         ?session_id muAccount:impersonates ?maybe_impersonated.
-
-         ?maybe_impersonated a foaf:OnlineAccount;
-           ext:sessionRole ?session_role;
-           foaf:accountServiceHomepage <https://github.com/lblod/mock-login-service>.
-
-         ?person a foaf:Person;
-           foaf:account ?maybe_impersonated;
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -8,7 +8,7 @@ x-logging: &default-logging
 
 services:
   loket:
-    image: lblod/frontend-loket:0.92.1
+    image: lblod/frontend-loket:0.92.3
```
Windvis, line 11: This needs to be updated once we bundle the frontend changes.

config/authorization/config.ex
```diff
@@ -123,6 +94,31 @@ defmodule Acl.UserGroups.Config do
...
+          }
+
+          {
+            ?session_id ext:sessionRole ?session_role .
+          } UNION {
+            ?session_id ext:originalSessionRole ?session_role .
+          }
+          BIND(COALESCE(?maybe_original_session_role, ?current_session_role) AS ?session_role)
+        }
+        LIMIT 1"
```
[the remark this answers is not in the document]
Windvis: I think we can simply remove the `BIND(COALESCE( ...` since it references things that don't exist anymore after I did some renaming. 

I don't understand why the query doesn't error though, since the spec says that it should if both variables are unbound: https://www.w3.org/TR/sparql11-query/#func-coalesce

> If none of the arguments evaluates to an RDF term, an error is raised. If no expressions are evaluated without error, an error is raised.

docker-compose.yml
```diff
@@ -8,7 +8,7 @@ x-logging: &default-logging
 
 services:
   loket:
-    image: lblod/frontend-loket:0.92.1
+    image: lblod/frontend-loket:feature-impersonation-refactor
```
Windvis, line 11: I used a feature image for easier testing, we should replace it with a tagged release before merging this.

config/authorization/config.ex
```diff
@@ -22,44 +22,15 @@ defmodule Acl.UserGroups.Config do
...
-      }
-      LIMIT 1
-    """
+    "PREFIX ext: <http://mu.semte.ch/vocabularies/ext/>
+     PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
+     SELECT DISTINCT ?session_group ?session_role WHERE {
+      <SESSION_ID> ext:sessionGroup/mu:uuid ?session_group;
+                   ext:sessionRole ?session_role.
+      FILTER( ?session_role = \"#{group_string}\" )
+    }"
```
Windvis, line 31: (This simply undoes the changes made in feature/person-impersonation-service. So back to the state in dev/master)

Riadabd (approved):
All looks great :+1: Tested multiple flows, and there were no issues.

config/authorization/config.ex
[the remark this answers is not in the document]
Riadabd: I wanted to check this out of curiosity, and it seems it errors out when running it directly in the virtuoso sparql interface. This line has been removed now, but it's good to note in the future (maybe `mu-auth` does some re-writing before executing the query which caused the difference in behavior?).

PR #566, opened 2024-04-26
Chore/use updated delta publication graph maintainer

cecemel:
This PR, uses the version of publication graph maintainer: https://github.com/lblod/delta-producer-publication-graph-maintainer/pull/32
Updated the configuration. (Even if not strictly necessary)

If you want to test `drc up -d`. Assuming you ran all initial syncs. Else you can  use the configuration `waitForInitialSync: false` per deltastream.

What to test?
- does it produces delta files at the expected location. (So just check the created files)
- Doe it still respond to background jobs?

I don't expect any other issues.
SeeAlso: DL-4527

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
 # Changelog
+## Unreleased
+### Fixes
+- Bump delta-producer-publication-graph-maintainer [DL-4527] and related [OP-3151]
+### Deploy Notes
+- `drc up -d delta-producer-publication-graph-maintainer; drc restart deltanotifier`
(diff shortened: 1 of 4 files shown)


Riadabd (approved):
Tested some flows, and it looks good to go :+1: 
* Delta files get correctly generated.
* Service runs in the background as expected.

I also ran `app-worship-decisions-database` to consume some delta files, and the flow worked as expected:
* `submissions-consumer` successfully consumed worship submissions from Loket.
* `files-consumer` correctly consumed the files as well.
* `delta-producer-publication-graph-maintainer` logs correctly showed the service being polled for new deltas.

PR #584, opened 2024-07-02
[DL-6020] Remove outdated email predicates

Windvis:
These predicates are part of the berichtencentrum module, but due to some previous mu-auth config (or migrations) the predicates also ended up in different graphs. This wasn't really an issue until the values started to get out of sync since mu-cl-resources then returns the wrong value sometimes.

By removing the predicates from all graphs besides the berichtencentrum one we fix the problem.


Riadabd (approved):
Tested locally, and all triples belonging to non-berichten graphs were deleted as expected.

PR #594, opened 2024-08-22
Adjust download/berichtencentrum url warning cleanup job

Riadabd:
## PR Description

This PR ratifies some extra deletion that happend with the previous version of this cleanup job where result containers were deleted (these containers are needed for the `download-url` service to query failed downloads).

The PR also adds a dummy container to restore the link for failed remote data objects (i.e., ones with `<http://lblod.data.gift/file-download-statuses/failure>` status).


cecemel (approved):
ok! thanks for that. Maybe better to target the master branch and then merge back in development so we can quickly deploy this.
Also, can you make a ticket for this and add this to this sprint.

PR #596, opened 2024-08-26
Fix Submission export filters

benjay10:
**[DL-6084] [DL-6085] [DL-6078] [DL-6079]**

These filters now better filter out the types of documents that don't need to be exported. Previously, documents that have multiple values for `dct:type` would always be exported because of the way SPARQL does pattern matching, no matter of their filtered document types.

**What went wrong (crude example):**

Very often, FormData for a Submission has triples like the following:

```turtle
<someFormData>
  ...
  dct:type foaf:Document ;
  dct:type besluitType:abcde-12345 .
```

The filter used in the config was something like:

```SPARQL
<someFormData> dct:type ?besluitType .
FILTER (?besluitType NOT IN (<notAllowed1> <notAllowed2> ...))
```

But this would in all cases still allow subject `<someFormData>` to be exported. In the query above, `?besluitType` can be pattern matched to `foaf:Document`, which is not in the filter to be ignored, and thus this subject appears in the results of the query.

**How it was fixed:**

Instead of using a `FILTER - NOT IN` we used a `FILTER NOT EXISTS` to make sure no triple for a besluit type that is not allowed can be found. E.g.:

```SPARQL
FILTER NOT EXISTS { <someFormData> dct:type <notAllowed1> . }
FILTER NOT EXISTS { <someFormData> dct:type <notAllowed2> . }
```

**TO DO / TO DISCUSS:**

* These filters where are applied to the Submission and the SubmissionDocument. This means that subjects for FormData, FileDataObject, RemoteFileDataObject, ... are still exported, even if they belong to Submissions that are not allowed to be exported. **Should we add the same filter to all related subjects?**
* This config is not readable and is difficult to maintain. Because the config is in JSON, there can be no newlines (only `\n` escape characters). The same filter is also repeated for every relevant subject and there is no way to see if those filters are still the same. If the filter needs to be changed, we need to changed it everywhere in this document, but they sometimes need to be slightly different. **Can we change this to pure JavaScript so we can use variables to compose the filters to avoid this repetition?**


Riadabd:
> Should we add the same filter to all related subjects?

I think so, but it would be best for Felix to confirm.

> Can we change this to pure JavaScript so we can use variables to compose the filters to avoid this repetition?

I would personally like if it can be transformed. As it stands, the configuration is quite unreadable (because of needing to add `\n` for newlines). The config requires adjustment in production from time to time when exports to `app-toezicht-abb` fail (in this case, we end up updating the `FILTER` clause for the date) and having it as JSON means it is inconvenient to read and adjust.

This does mean that the service itself needs to be modernized, which would be a nice side-effect. It was last released 3 years ago, and we are currently running a hotfix branch (`lblod/mandaten-download-generator-service:0.3.6-hotfix_4`) in Loket.

benjay10:
@Riadabd 

Felix fears that making modifications to the service would be too time consuming at the moment and that adding the filters to other subjects is best left to a different ticket.

Riadabd (approved):

PR #606, opened 2024-10-17
Extend op consumer to all unit types

claire-lovisa:
# Context

DL-6217 & DL-6218

# Details

In this PR, we:
1. Flush all data that is going to be re-imported by the consumer as well as the sync jobs
2. We extend the rules so that consumed administrative units have all they need (concept schemes mostly). An extensive analysis of what needs to be added can be found in the comments of DL-6218.

# How to test

@cecemel said it all in his analysis on DL-5713:
```
How I would test: first create a bestuurseenheid in OP and see whether (after a while) you can:
    - CRUD an inzending voor toezicht
    - CRUD a bericht
    - CRUD BBCD
    - CRUD leidinggevenden
    - CRUD personeelsbeheer (if you fixe the migration)
    - CRUD manndatarissen, potentiall not needed as the master-app is going to be LMB, but still a good sanitycheck. Mind this is only for political stuff.
```

I have tested for PEVA Gemeente, it flow to Loket, the `update-bestuurseenheid-mock-login` service does create the mock-login user and we can access the app.

# What's next

DL-6219: we should now create `Personeelsdatabank` data for those imported admin units. The mock-login service still does its job as long as we don't consume other organizations that are not `besluit:Bestuurseenheid`.


cecemel:
Thanks, so far it's progressing, but some bugs pop up during testing.
# Too many bestuursorganen in toezicht.
It's seems we need a more fine grained mapping rule for bestuursorganen to add them to specific concept:schemes. 
Now we have e.g. the situation:
image
Here it's clear we shouldn't see the leidinggevenden organen.

# Too few bestuursorganen get exported by OP.
E.g. for https://data.lblod.info/id/bestuurseenheden/2c80d772-5a21-4e9d-be8a-35e9bb6d66c6
We expect 4 bestuursorganen. We only get 1. That's an issue at the OP side, not all bestuursorganenclass codes were added to
the export concept scheme.
```
SELECT DISTINCT ?s ?label {

?s ?in <http://lblod.data.gift/concept-schemes/10caf4a4-1ff5-4c05-917e-56489910d68b>.

?s skos:prefLabel ?label

}
```

cecemel (approved):

PR #607, opened 2024-10-25
Cleanup job: personeel data

claire-lovisa:
# Context

DL-6219

We want to add data structure to fill in personel information for new administrative units

# Todo

1. We are waiting for the list of administrative unit types that should be allowed/required to fill in personel information. A comment has been left on the ticket, waiting for feedback
2. The current query seems to big to come through. It might be solved if we filter out more classifications. If not, we'll have to find a trick to make it work, maybe by decomposing into multiple queries (one creating datasets, one creating slices, one creating observations for example)


cecemel (approved):
Thanks! I changed some things. See commits.

PR #609, opened 2024-10-30
fix: delete double mandatories and ministers for a worship service

mirdono:
The mandatories and ministers were entered twice for a worship service. Once via
Loket and later via vendor software. Since this worship service is currently
managed via the vendor software, this migration removes the deprecated entries
initially entered in Loket.

## Related ticket
- DL-6188


Riadabd:
Thank you @mirdono! Migration was run directly on production.

PR #611, opened 2024-11-06
[DL-6231]: fusies 2025 migrations

wolfderechter:
## ID

DL-6231

## Description

Add the new fuesies of 2025.

Used the migrations from GN, since they last added it: https://github.com/lblod/app-gelinkt-notuleren/pull/201/files


Riadabd:
Thanks @wolfderechter !

PR will be closed as migrations will be run directly on production.

PR #624, opened 2024-12-17
Export `betrokken lokale besturen` with submissions

claire-lovisa:
This is information that is now required by Toezicht ABB to be able to show properly cross-referenced documents.

# Testing

This can be tested together with https://github.com/lblod/app-toezicht-abb/pull/48, with the same instructions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## Unreleased
+
+- Export betrokken lokale besturen with submissions [#624]
+
+### Deploy notes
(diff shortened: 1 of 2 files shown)


benjay10 (changes requested):
After weeks of on and off debugging I finally found some config errors on my side. Submissions are now shown in Toezicht ABB on my machine, but the referenced submissions where still missing. I added the Artikel entitiy and its relations, and a `dct:relation` from the submission document to complete the cross referencing data in the export.
Could you take a look to see if that is correct?

PR #629, opened 2025-01-15
Migration: query to remove data with language tag

benjay10:
**[DL-6301]** (and **[DL-6286]**)

Remove only harvested data that with a language tag is a duplicate of already existing data without a language tag. These tags remained after the producer-consumer weren't able to remove this.

The hope is that the producer in Loket for OP is going to heal and see that these triples also need to be removed in OP.


cecemel (commented):
Given the changes in the consumer, wouldn't it be simpler to flush all data here and sync from scratch?

benjay10:
It would indeed make a lot of sense. I have changed this PR to a full flush and resync.

cecemel:
> Also: how about the producer making deltas during the re-sync? 
Yeah maybe better turn it off during the ingest. To not overload the system. (I would expect it works in theory)

> so the producer is going to have to be reset.
It think by flushing the consumer side, the 'ground truth' (i.e. the source in loket for the producer) will get some updates, and the healing side of the producer should take care of making sure everything.

We did a similar operation for 'normal' mandatarissen and LMB, where we also flushed the 'source data', ingested again, and re-produced the data for other consumers. That worked,

cecemel (approved):

PR #632, opened 2025-01-31
Re-export missing submissions

benjay10:
**[DL-6349]**

A number of submissions were missed by the `prepare-submissions-for-export` service, certain submissions were not properly exported by the `delta-producer-publication-graph-maintainer`, and certain submissions were not properly consumed and/or dispatched by the Worship Decisions Database application. Solved by re-exporting all of these by two mechanisms:

* Some submissions still needed their `schema:publication` triple
  - Sending mock delta messages via a script to the `prepare-submissions-for-export` service to restart the process 
* The rest is simply removed from the `publication-triplestore` and re-produced via healing

These instructions are laid out in the CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,43 @@
 # Changelog
 
+## Unreleased
+
+### General
+
+- Add `migrations-publication-triplestore` as a migration service for the publication-triplestore [DL-6349]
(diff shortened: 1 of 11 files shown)


claire-lovisa (approved):
The instructions are clear, I tested and was able to see the missing submissions being produced. It's also handy to have a migration service for the publication triplestore, will help keeping trace of the fixes we do

PR #635, opened 2025-02-05
Add report for duplicate harvested and non harvested mandatarissen

Riadabd:
## Ticket ID

DL-6342

## Description

Organizations moving from Loket to a vendor (or vice-versa) may end up having duplicate mandatarissen, which confuse the user. The report will serve as a sentinel to catch any future cases.

## How to test

1. `docker compose restart report-generation` to load the report
2. `docker compose exec report-generation curl -X POST -H "Content-Type: application/vnd.api+json"  -d '{  "data": { "attributes": { "reportName": "harvested-and-non-harvested-mandatarissen-report" }  }  }' http://localhost/reports` to run the report.
3. If you are running it on a recent production backup, it should a count of around 464 rows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Add `migrations-publication-triplestore` as a migration service for the publication-triplestore [DL-6349]
 - Re-export 12 submissions and their related subjects, because they are missing from Worship Decisions Database, see migrations and deploy instructions [DL-6349]
+- Add report to detect harvested and non-harvested mandatarissen having the same position in the same organization. [DL-6342]
 
 ### Deploy instructions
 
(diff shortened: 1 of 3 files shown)


benjay10:
I was able to test this and it technically works correctly.

Some remarks:

* In the query, you match for the triples `?timedGoverningBody <http://data.vlaanderen.be/ns/mandaat#bindingStart> ?startDate .`, but that is not used elsewhere. Is this intentional? Does 
the body always need their start date set?
* The report shows the columns `subject` and `subjectHarvested`. I think these names can be improved to, e.g. `Mandate position (entered in Loket)` and `Mandate position (via harvesting)`.
* Maybe useful to also include the URI of the governing body, or the URI of the mandate that has double positions.

Riadabd:
> I was able to test this and it technically works correctly.
> 
> Some remarks:
> 
>     * In the query, you match for the triples `?timedGoverningBody <http://data.vlaanderen.be/ns/mandaat#bindingStart> ?startDate .`, but that is not used elsewhere. Is this intentional? Does
>       the body always need their start date set?
> 
>     * The report shows the columns `subject` and `subjectHarvested`. I think these names can be improved to, e.g. `Mandate position (entered in Loket)` and `Mandate position (via harvesting)`.
> 
>     * Maybe useful to also include the URI of the governing body, or the URI of the mandate that has double positions.

* Indeed, `?startDate` was a remnant of data debugging. I removed it.
* The column names have been updated to `mandatePositionLoket` and `mandatePositionHarvested` respectively.
* I added the worship unit URI and the governing body post URI for extra visibility into the data.

benjay10 (approved):
Yes, looks good to me!

PR #638, opened 2025-02-11
Update mandatarissen producer to stop producing what OP produces

claire-lovisa:
:warning: DO NOT MERGE BEFORE ALL KNOWN CONSUMERS HAVE BEEN UPDATED

# Description

**DL-6210**

We are going to add the op-public-consumer to app-mandatendatabank, causing some triples being exported from Loket to MDB to be obsolete. In this PR, we:
1. Remove parts of the configuration of the mandatarissen producer so that the triples coming from OP stop being produced by Loket
2. Remove the same triples from the mandatarissen producer graph. This is to prevent annoying side effects as other applications consume from Loket and OP, for example it prevents having some administrative units being deleted from consuming applications.

# Testing instructions

1. Get a working Loket stack with some data in it. I tried on PROD data, maybe you can try with DEV or QA data, but whatever is the easiest for you
2. Before checkout-ing this branch, run a healing to be sure to be up-to-date on your publication triplestore: `drc exec delta-producer-background-jobs-initiator curl -X POST http://localhost/mandatarissen/healing-jobs`
3. Also run a dump `drc exec delta-producer-background-jobs-initiator curl -X POST http://localhost/mandatarissen/dump-publication-graph-jobs`
4. Checkout this branch and `drc restart migrations-publication-triplestore delta-producer-publication-graph-maintainer`
5. When the migrations are done running, run again a healing and when the healing is done a dump

Expected result:
- After the healing, no deltas should have been produced about the resources removed from the configuration
- The more recent dump should be substantially smaller and should not contain removed data

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Remove submissions export config replaced by op sync [DL-6394]
 - Repair old cross referencing submissions from CKB's. These predate the cross referencing feature. [DL-6415]
 - Add migration that removes three older, broken EredienstMandatarissen pointing to non-existing people and contact points [DL-5662]
+- Update mandatarissen producer to stop producing what OP produces [DL-6210]
 
 ### Deploy instructions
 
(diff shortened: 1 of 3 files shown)


cecemel (approved):

cecemel:
@benjay10 @claire-lovisa I approved this one. But we will only merge once all consumers are updated. So putting this on draft. 
Note: I do find things without subject, but I think this is related to the implementation problems we discussed. (And unrelated to this PR)

cecemel (approved):
@claire-lovisa @benjay10 this one is good. I will deploy it on dev.
We go ahead as discussed in chat.

PR #642, opened 2025-02-17
Remove fusiegemeenten for mock-login healing

benjay10:
**[DL-6375]**

Certain fusiegemeenten don't have the correct, updated, name on the mock-login or impersonation page. This PR adds a migration to remove mock-login accounts for those fusiegemeenten and their OCMW. With the healing on the `update-bestuurseenheid-mock-login-service`, these are added again, but this time with the correct labels.


aatauil (approved):
This indeed removed the mock-login accounts

PR #644, opened 2025-02-21
Prune leidingevenden producer to stop producing what OP produces

benjay10:
**[DL-6449]**

See https://github.com/lblod/app-digitaal-loket/pull/638

The notes that would come in this description would be almost identical to the ones in the linked PR above from @claire-lovisa. Please use those notes and apply them to the `app-leidingevenden-databank`.

**NOTE:** I copied what @claire-lovisa did in their PR, but applied those changes to leidinggevenden.

**NOTE:** I found some predicates that seem to be added in Loket on the `Bestuursorgaan` and `Bestuursfunctie`. See the migration on the publication triplestore, where I attempt to keep them. I assume these are important.

**TODO:** I would like to still run a test, such as in the aforementioned PR to see if all runs well.

### DIFF
--- a/config/migrations-publication-triplestore/2025/20250221050000-flush-leidingevenden-triples-in-favour-of-op-public-sync.sparql
+++ b/config/migrations-publication-triplestore/2025/20250221050000-flush-leidingevenden-triples-in-favour-of-op-public-sync.sparql
@@ -13,7 +13,7 @@ WHERE {
       <http://mu.semte.ch/vocabularies/ext/BestuurseenheidClassificatieCode>
     }
     ?s
-      rfd:type ?type ;
+      rdf:type ?type ;
       ?p ?o .
   }
 }
@@ -28,7 +28,7 @@ DELETE {
 WHERE {
   GRAPH <http://redpencil.data.gift/id/deltas/producer/loket-leidinggevenden-producer> {
     ?s
-      rfd:type <http://data.vlaanderen.be/ns/besluit#Bestuursorgaan> ;
+      rdf:type <http://data.vlaanderen.be/ns/besluit#Bestuursorgaan> ;
       ?p ?o .
 
     FILTER (?p NOT IN (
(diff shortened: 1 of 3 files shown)


config/migrations-publication-triplestore/2025/20250221050000-flush-leidingevenden-triples-in-favour-of-op-public-sync.sparql
```diff
@@ -0,0 +1,60 @@
...
+}
+WHERE {
+  GRAPH <http://redpencil.data.gift/id/deltas/producer/loket-leidinggevenden-producer> {
+    VALUES ?type {
+      <http://mu.semte.ch/vocabularies/ext/BestuursorgaanClassificatieCode>
+      <http://data.vlaanderen.be/ns/besluit#Bestuurseenheid>
+      <http://mu.semte.ch/vocabularies/ext/BestuurseenheidClassificatieCode>
+    }
+    ?s
+      rfd:type ?type ;
```
claire-lovisa, line 16: There is a typo, `rfd` should be `rdf`

config/migrations-publication-triplestore/2025/20250221050000-flush-leidingevenden-triples-in-favour-of-op-public-sync.sparql
```diff
@@ -0,0 +1,60 @@
...
+
+DELETE {
+  GRAPH <http://redpencil.data.gift/id/deltas/producer/loket-leidinggevenden-producer> {
+    ?s ?p ?o .
+  }
+}
+WHERE {
+  GRAPH <http://redpencil.data.gift/id/deltas/producer/loket-leidinggevenden-producer> {
+    ?s
+      rfd:type <http://data.lblod.info/vocabularies/leidinggevenden/Bestuursfunctie> ;
```
claire-lovisa, line 51: And here as well

claire-lovisa (changes requested):
When running the healing after the flush + update of the configuration I noticed that some organen types were deleted by the producer. It's an issue because it means some consuming apps will have organen without a type ahah. I think it's because when the orgaan doesn't have a `<http://data.lblod.info/vocabularies/leidinggevenden/heeftBestuursfunctie>` triple, the type is not produced at all. So when flushing, we should only keep the type if we have a `heeftBestuursfunctie` predicate on the orgaan. This is just a supposition, to be tested!

claire-lovisa (approved):
I added a commit to ensure the leidinggevenden producer will not produce any data after the update in the configuration. Feel free to take a second look, but I think now the PR is ready!

PR #652, opened 2025-03-16
Opt out of VDDS deltas for `resource` and `toezicht-flattened`

Riadabd:
## Description

The PR adds an VDDS deltas opt outs for `resource` and `toezicht-flattened-form-data-generator`.

* **resource**: This tracks what we already do for `app-worship-decisions-database`. There's no need for `resource` to be aware of VDDS deltas.
* **toezicht-flattened-form-data-generator**: This was part of [this commit for `app-meldingsplichtige-api`](https://github.com/lblod/app-meldingsplichtige-api/commit/d89cf5e6467b1c4a4dcccd162cba8a3199bc8f33). VDDS for Loket matches on `http://www.w3.org/ns/adms#status` and may insert documents that have a `adms:status <http://lblod.data.gift/concepts/9bd8d86d-bb10-4456-a84e-91e9507c374c>` that match the `toezicht-flattened` delta rules => `toezicht-flattened` will then receive a delta match which will re-trigger a flattening on that submission (which may have already occurred in the past).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Update leidinggevenden producer to stop producing what OP produces [DL-6449]
 - Bump deliver-bbcdr [DL-6481]
 - Add missing `labels` key for some services. [DL-6490]
+- Opt out of VDDS deltas for `resource` and `toezicht-flattened-form-data-generator`.
 
 ### Deploy instructions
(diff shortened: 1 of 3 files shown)


benjay10 (approved):
This makes a lot of sense!

PR #653, opened 2025-03-18
Move consumers to delta-consumers folder

claire-lovisa:
To keep all the apps in the ecosystem consistent, we move all the consumers to a folder named `delta-consumers`. This should have no impact on the stack, a simple `drc up -d` picks up the new configuration location.


Riadabd (approved):
Ran the consumers locally, and they started up without issues.

PR #662, opened 2025-06-04
Frontends via dispatcher and ACM

benjay10:
The goal is to unify access to the frontends through the identifier and dispatcher, and their connection to each individual login service for ACM. This is needed because each login service has their own unique key.

This then has to be deployed with multiple virtual hosts on the identifier, corresponding to each frontend. All "controle" related services can then be removed.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,7 +8,7 @@ x-logging: &default-logging
 
 services:
   loket:
-    image: lblod/frontend-loket:0.100.0
+    image: lblod/frontend-loket:0.100.1
     links:
       - identifier:backend
     labels:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### General
 
+- Bump frontend updating privacy policy disclaimer [DL-5635]
 - Bump dashboard frontend to `v1.8.0`. [DL-6588]
   - This includes the ACM and impersonation changes.
 - Bump dashboard frontend to `v1.9.0`
@@ -18,6 +19,12 @@ to unify the login services, and prepare for vendor management
   JSON response on browser HTML requests.
 
 ### Deploy instructions
+
+**New version of Loket**
+
+```
+drc up -d loket
+```
 
 **ACM/IDM updates**
 


docker-compose.yml
```diff
@@ -137,6 +137,30 @@ services:
       - "logging=true"
     restart: always
     logging: *default-logging
+  login-dashboard:
+    image: lblod/acmidm-login-service:0.11.0
```
Windvis, line 141: There is a 0.12.0 release now as well, so I would use that (same for the others).

docker-compose.yml
```diff
@@ -137,6 +137,30 @@ services:
...
+    environment:
+      MU_APPLICATION_AUTH_DISCOVERY_URL: "https://authenticatie-ti.vlaanderen.be/op"
+      MU_APPLICATION_AUTH_CLIENT_ID: "a2c0d6ea-01b4-4f68-920b-10834a943c27"
```
Windvis, line 144: I would use the correct ids already, if we have it. Or snip it out so we just configure it on the server. If anything, we should probably add the prod values, but we don't do it for the loket version either.

docker-compose.yml
```diff
@@ -137,6 +137,30 @@ services:
...
+      MU_APPLICATION_AUTH_REDIRECT_URI: "https://dashboard.lblod.info/authorization/callback"
+      MU_APPLICATION_AUTH_CLIENT_SECRET: "secret"
+      LOG_SINK_URL: "http://sink"
+    labels:
+      - "logging=true"
+    restart: always
+    logging: *default-logging
+  login-vendor-management:
```
Windvis, line 152: Did you add this in preparation for the ACM/IDM setup in the dashboard or something? :smile:

config/dispatcher/dispatcher.ex
```diff
@@ -650,14 +652,44 @@ defmodule Dispatcher do
...
+  end
+
+  get "/@appuniversum/*path", %{ reverse_host: ["loket" | _rest] } do
+    forward conn, path, "http://loket/@appuniversum/"
+  end
+
+  match "/*_path", %{ reverse_host: ["loket" | _rest] } do
+    # *_path allows a path to be supplied, but will not yield
+    # an error that we don't use the path variable.
+    forward conn, [], "http://loket/index.html"
```
Windvis, line 678: Does this actually work already? The loket frontend still uses the old proxy-service based setup which expects to be the entry point.

docker-compose.yml
benjay10: For the vendor-management, yes. Not good? One PR for everything?

docker-compose.yml
benjay10: I just copied whatever was set on the login service for loket.

I think the reasoning here is that it is okay to share these values from the QA setup, but not the ones from the production setup. These QA values now work as a nice example of how to setup the overrides for production.

config/dispatcher/dispatcher.ex
```diff
@@ -650,14 +652,44 @@ defmodule Dispatcher do
     forward conn, path, "http://berichtencentrum-melding/melding"
   end
 
+  #################################################################
+  # Loket
+  #################################################################
+
+  # Login
+
+  match "/sessions/*path", %{ reverse_host: ["loket" | _rest] } do
```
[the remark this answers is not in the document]
benjay10: But then we end up with dispatcher rules that have overlap and where ordering matters. The rules are thus not mutually exclusive / not logically separated. Can we add additional new rules instead like the following?

```elixir
match "/sessions/*path", %{ reverse_host: ["dev", "loket" | _rest] } do
   [...]
```

config/dispatcher/dispatcher.ex
benjay10: Yes, this fully works. At least for me locally. The backend on the proxy service references the identifier. I assume the proxy service then just forwards requests exactly like they come in, virtual host information included, but the dispatcher will look at the MIME types to correctly forward all requests.

docker-compose.yml
Windvis: > but not the ones from the production setup.

I don't think that's true, since everything is public besides the client secret. I guess it's just done this way because we usually don't have the prod values yet when we merge / deploy.

I don't mind keeping QA values, but it feels strange to keep the _wrong_ ones in there. We could do something like in LPDC if we don't know the values?

https://github.com/lblod/app-lpdc-digitaal-loket/blob/5e85eb2ec37ce1879b5b35f312e314b021e12a17/docker-compose.yml#L145-L150

config/dispatcher/dispatcher.ex
Windvis: Fair enough :+1: 

We should probably just change the base image in the frontend though, since it's a quick change and we were planning on doing that anyways. I'll create a PR.

docker-compose.yml
benjay10: Okay, I'll use the production values for all three. That would make the deploy a little easier. Except for the secret though, but I'll make it a little easier to see that it needs to be replaced.

config/dispatcher/dispatcher.ex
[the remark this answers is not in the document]
benjay10: Okay, I'll make use of the ordering, so that it catches all attempts for frontend access.

docker-compose.yml
```diff
@@ -125,13 +125,37 @@ services:
...
+    labels:
+      - "logging=true"
+    restart: always
+    logging: *default-logging
+  login-dashboard:
+    image: lblod/acmidm-login-service:0.12.0
+    environment:
+      MU_APPLICATION_AUTH_DISCOVERY_URL: "https://authenticatie.vlaanderen.be/op/.well-known/openid-configuration"
+      MU_APPLICATION_AUTH_CLIENT_ID: "90a39574-e986-4007-84f2-becf6d9eb481"
+      MU_APPLICATION_AUTH_REDIRECT_URI: "https://loket.lokaalbestuur.vlaanderen.be/authorization/callback"
```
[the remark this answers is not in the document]
benjay10: What? I copied these from production?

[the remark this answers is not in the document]
Windvis: But this is the config for the dashboard login? Not the Loket one? Do we already have the prod values for the dashboard?

Windvis (approved):
Looks good. Could you add some a changelog entry and some deploy instructions?

PR #672, opened 2025-07-23
Reorganize consumer configuration files

claire-lovisa:
To ease debugging, I reorganized the consumer files following a suggestion from Felix in #669 .

The content of the rules has been just slightly split sometimes to ease the naming of the files, but it should not affect how it works.


cecemel (approved):
Hey this looks good.
I think ideally it should be deployed on DEV and kevin should do CRUD of all affected resource types. 
thanks!

PR #676, opened 2025-09-01
Feature/sparql parser addendums

cecemel:
A couple of addendums/corrections to https://github.com/lblod/app-digitaal-loket/pull/674


cecemel:
A couple of addendums/corrections to https://github.com/lblod/app-digitaal-loket/pull/674

PR #679, opened 2025-09-03
[DL-6543] check vendor API key against database hash

kobemertens:
- **add initial config.lisp**
- **add grants**
- **add sparql-parser config**
- **change database to sparql-parser service**
- **no more variables, use plain strings instead**
- **add some comments for review**
- **remove key access from auth config**
- **remove vendor key from mu-cl-resource**
- **remove deprecated version number from docker-compose.yml**
- **fix mu-auth config**
- **update readme**


claire-lovisa (commented):
I tested the PR and it works properly. I'm not approving per say because I think that we have some TODOs before merging, mostly bumping `vendor-login`, `automatic-submission` and `berichtencentrum-melding`!

PR #730, opened 2026-04-17
Cross Referencing rules in triplestore

benjay10:
**[DL-7202] [DL-7204] [DL-7325]**

Adding the Cross Referencing rules (mappings) in data form as a migration onto the triplestore, also adding the predicates ready for export.


cecemel (approved):
Side notes, the ext name space is a bit muh for vendors. But I guess changing that, will be too much work.

cecemel (approved):

PR #732, opened 2026-04-27
[DL-6577]: switch to sparql parser

wolfderechter:
## ID

DL-6577

## Description

Switch to sparql parser

I've checked the different modules in loket, and everything seems to work as expected

Continues the work of https://github.com/lblod/app-digitaal-loket/pull/695


NikiDeVisscher (commented):
The modules work, but I have issues when trying to access the /search route. The first time, it takes several minutes to load. Once I notice from the logs that all the data is loaded and I refresh, the page works just fine.

wolfderechter:
> The modules work, but I have issues when trying to access the /search route. The first time, it takes several minutes to load. Once I notice from the logs that all the data is loaded and I refresh, the page works just fine.

might be the elasticsearch index/ipdc consumer?

aatauil:
@wolfderechter Went through the app. Works as I expected. I am not completely up to date with the state of new loket. So discovered berichtencentrum and went through old modules as I know them from old loket. App reacts faster, Crudded everything there is to crud. Only existing toezicht dossiers I couldnt open because my machine wont handle downloading all toezicht files.

aatauil (approved):

PR #737, opened 2026-05-19
Hotfix/fix migration gemeentewegen

cecemel:
We had a problem with performance with the previous migrations. So we had to re-write them. This commit splits the mega-migration in smaller chunks and filters on the specific type of submission. See also [DL-7397]


NikiDeVisscher (approved):
Tested locally on (old) prod data: all migrations run successfully

PR #742, opened 2026-06-04
DL-7401: Make ldes-graph not public

kikef98:
Restrict access to the ipdc-ldes graph by removing it from the public grant and limiting it to logged-in/impersonating.

### Testing
Verify that:

1. Unauthenticated requests no longer returns data
2. Authenticated requests still returns data as expected

Note: Felix said re-index might be needed but when I tested locally it worked without having to do this, after restarting the services. 

### Ticket
DL-7401

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,10 @@
 ```
 drc restart database identifier
 ```
-
+Then search should be re-indexed
+```
+mu script search manage-indexes # follow the steps there. Re-index "public-services"
+```
 # v1.223.1
 - Fix dispatcher rules. Postfix: [DL-7316].
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 # Unreleased
 
 - Add Cross Referencing rules on the BesluitType and BesluitDocumentType [DL-7202] [DL-7204] [DL-7325]
+- Make ldes-graph not public [DL-7401]
+
+## Deploy notes
+```
+drc restart database identifier
+```
 
 # v1.223.1
 - Fix dispatcher rules. Postfix: [DL-7316].
--- a/config/authorization/config.ex
+++ b/config/authorization/config.ex
@@ -215,23 +215,7 @@ defmodule Acl.UserGroups.Config do
                     graph: "http://mu.semte.ch/graphs/sessions",
                     constraint: %ResourceFormatConstraint{
                       resource_prefix: "http://mu.semte.ch/sessions/"
-                    } },
-                  %GraphSpec{
-                    graph: "http://mu.semte.ch/graphs/ipdc/ldes-data",
-                    constraint: %ResourceConstraint{
-                      resource_types: [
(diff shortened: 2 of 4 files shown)


config/authorization/config.ex
```diff
@@ -280,7 +264,23 @@ defmodule Acl.UserGroups.Config do
                         "http://xmlns.com/foaf/0.1/Person",
                         "http://xmlns.com/foaf/0.1/OnlineAccount",
                         "http://www.w3.org/ns/adms#Identifier",
-                      ] } } ] },
+                      ] } },
+                  %GraphSpec{
+                    graph: "http://mu.semte.ch/graphs/ipdc/ldes-data",
```
cecemel: I think it's better here: https://github.com/lblod/app-digitaal-loket/blob/e7b6b01690d2fccc0cc68a37c5cd8a79a23bd798/config/authorization/config.ex#L222

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 # Unreleased
 
 - Add Cross Referencing rules on the BesluitType and BesluitDocumentType [DL-7202] [DL-7204] [DL-7325]
+- Make ldes-graph not public [DL-7401]
+
+## Deploy notes
+```
+drc restart database identifier
```
cecemel, line 10: no re-index?

kikef98: When I tested locally it was not necessary, worked without., just by running the changelog commands mentioned here. I added this in the note in this PR. Should I still add the re-index?

config/authorization/config.ex
```diff
@@ -215,7 +215,21 @@ defmodule Acl.UserGroups.Config do
                     graph: "http://mu.semte.ch/graphs/sessions",
                     constraint: %ResourceFormatConstraint{
                       resource_prefix: "http://mu.semte.ch/sessions/"
-                    } },
+                    } }] },
+      %GroupSpec{
+        name: "public-r",
```
cecemel, line 220: I think it's best to rename it to "ipdc-ldes" and  create a separate groupSpec for that. Als mu-search is going to be weird.
I also forgot to specify in needs to be specified in search config as extra group
```
  "eager_indexing_groups": [
    [
      {
        "name": "public",
        "variables": []
      }
    ]
  ]
So an extra group here, that matches groupspec

File: CHANGELOG.md (hunk 2 above)
Comment (cecemel) [reply CHANGELOG.md:10]: It still needs a re-index I think. I also realize it's "workish" without re-index because the identifier needs extra config:
https://github.com/lblod/app-digitaal-loket/blob/master/docker-compose.yml#L40
```
environment:
   DEFAULT_MU_AUTH_ALLOWED_GROUPS_HEADER: '[{"variables":[],"name":"clean"}, {"variables":[],"name":"public"}, {"variables":[],"name":"lmb-public"}]'
````
can you add that too?

Comment (kikef98) [reply CHANGELOG.md:10]: Done, added both

Comment (cecemel) [reply CHANGELOG.md:10]: sorry @kikef98 I meant:
```
DEFAULT_MU_AUTH_ALLOWED_GROUPS_HEADER: "[{\"variables\":[],\"name\":\"clean\"}, {\"variables\":[],\"name\":\"public\"}]"

PR #745, opened 2026-06-17
DL-7423 - Missing submissions in erediensten: Migrations

kikef98:
There are some missing instances in erediensten. These migratiosn will make sure the healing picks them up

There are 3 types:
- Delete instance from publication-triplestore. It will be picked up again and go through flow
- 3 migrations (forms, submission and attachments) for instances without the publication flag at all. Here its added
- 1 migration for an instance where the publication flag was only missing on formData.

### To test:
Run the migration and see if the 3 things mentioned above have been done

### Ticket
DL-7423


cecemel (approved):
- ext:SubmissionDocument flag was missing
-  migration with same timestamp (better not to have it, even if here not a real problem)
- fixed

PR #746, opened 2026-06-24
Hotfix/sessions on vendor login

cecemel:
Hotfix; see `CHANGELOG.md`  [DL-7446]


CHANGELOG.md
```diff
@@ -1,5 +1,26 @@
 # Changelog
 
+# v1.224.2 (2026-06-24)
```
benjay10, line 32: A single PR with two separate release versions in it? I think in that case you don't have to split things up.

PR #747, opened 2026-06-26
VDDS: specific org graph for vendor data

benjay10:
Select vendor data from the specific organisation graph, instead of the full triplestore. This is needed after redistribution of vendor data in PR https://github.com/lblod/app-digitaal-loket/pull/741.

Without this change, the VDDS is really slow. The query for retrieving organisation and vendor data takes multiple seconds to finish. Healing would take days and regular processing would be unnecessarily heavy.

To test, use up to date test data, run healing on the VDDS and see if Submissions are really being processed. (Make sure you don't see the message

```
HEALING: No target graph found for subject <http://data.lblod.info/submissions/[a uuid...]>. Skipping.
```

too often in the logs. Some occurrences of this message could be normal.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+# Unreleased
+
+- VDDS: optimise query for vendor and organisation info. A needed speedup after re-distrubition of vendor info in [PR #741](https://github.com/lblod/app-digitaal-loket/pull/741)! No ticket.
+
 # v1.224.0 (2026-06-22)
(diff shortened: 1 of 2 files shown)


cecemel (commented):
Would this be really faster? 🤔   The vendor information also reside in a fixed graph, so maybe use that one instead?

Also, isn't it needed for all VDDS config, ie. <FailedessageHarvestJob>, etc..

cecemel (approved):

